PR #4, opened 2017-04-03
Fix wrong path

prumand:
The basic run task only work with the following adjustment of the path.


uglyog:
Thanks, I missed that when renaming the directories.

PR #30, opened 2017-12-08
Return a String instead of a serde_json value

elitau:
As this function is intended for use from other languages I would suggest to return a simple String instead of a serde_json::Value type.
Thus I don't need to have serde_json as a dependency in my [wrapper library](https://github.com/elitau/pact_elixir).


uglyog:
Good call, thank you for the PR

PR #32, opened 2018-02-12
fix json body matcher so root value strings and key value pairs are both valid

smckend:
When using JSON root value strings, the pact_matching library adds extra quotes to the string even if it already has them. 

Previous behaviour:
"example1" -> "\\\"example1\\\""
"\\\"example2\\\"" -> "\\\"\\\"example2\\"\\\""

This fix has just allowed both Strings and key-value HashMaps to pass as valid JSON.


uglyog:
Thanks for the PR

PR #42, opened 2019-07-28
Reqwest

### LINKED ISSUE #37: HTTPS pact broker not supported
Running the verifier with:

`./target/debug/pact_verifier_cli --provider-name myProvider --broker-url https://anything.com --hostname http://localhost:9999 --state-change-url http://localhost:9999/_pactSetup`

Yields:

```
Failures:

0) Failed to load pact - Could not load pacts from the pact broker 'https://anything.com' - IoError("Failed to access pact broker path \'/\' - \"Invalid scheme for Http\". URL: \'https://anything.com\'")
```

I guess this is related to https://github.com/hyperium/hyper/issues/1009, and https support needs to be provided as a separate library.

I would be more than happy to help the project, so maybe you'll see a PR from me when I get the time.

audunhalland:
Using `reqwest` for pact broker HTTP client.

Fixes #37

### DIFF
--- a/rust/pact_verifier/src/pact_broker.rs
+++ b/rust/pact_verifier/src/pact_broker.rs
@@ -35,10 +35,11 @@ fn content_type(response: &reqwest::async::Response) -> String {
 fn json_content_type(response: &reqwest::async::Response) -> bool {
     match content_type(response).parse::<mime::Mime>() {
         Ok(mime) => {
-            mime.type_() == mime::APPLICATION && (
-                mime.subtype() == "hal" ||
-                mime.subtype() == mime::JSON
-            )
+            match (mime.type_().as_str(), mime.subtype().as_str(), mime.suffix()) {
+                ("application", "json", None) => true,
+                ("application", "hal", Some(mime::JSON)) => true,
+                _ => false
+            }
         }
         Err(_) => false
     }
--- a/rust/pact_verifier/src/pact_broker.rs
+++ b/rust/pact_verifier/src/pact_broker.rs
@@ -466,11 +466,11 @@ mod tests {
 
         let client = HALClient::with_url(pact_broker.url().to_string());
         let result = runtime.block_on(client.clone().fetch(s!("/nonhal")));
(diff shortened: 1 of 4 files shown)


rust/pact_verifier/src/pact_broker.rs
```diff
@@ -31,24 +25,22 @@ fn as_string(json: &serde_json::Value) -> String {
...
-                Ok("application/hal+json") => true,
-                _ => false
-            }
-        },
-        None => false
+fn json_content_type(response: &reqwest::async::Response) -> bool {
+    match content_type(response).parse::<mime::Mime>() {
+        Ok(mime) => {
+            mime.type_() == mime::APPLICATION && (
+                mime.subtype() == "hal" ||
```
uglyog: I think the sub type here would be `hal+json`

audunhalland: I pushed a fix. With `mime` crate the `+json` part is the suffix.

rust/pact_verifier/src/pact_broker.rs
```diff
@@ -31,24 +25,23 @@ fn as_string(json: &serde_json::Value) -> String {
...
-            match value.to_str() {
-                Err(_) => false,
-                Ok("application/json") => true,
-                Ok("application/hal+json") => true,
+fn json_content_type(response: &reqwest::async::Response) -> bool {
+    match content_type(response).parse::<mime::Mime>() {
+        Ok(mime) => {
+            match (mime.type_().as_str(), mime.subtype().as_str(), mime.suffix()) {
+                ("application", "json", None) => true,
+                ("application", "hal", Some(mime::JSON)) => true,
```
audunhalland: There's a bug in `mime` that you cannot create a user-defined `mime::Name` (the return type of `type_`, `subtype` and `suffix`). You can only covert a name to a `&str` and not the other way around: https://github.com/hyperium/mime/issues/77. The bug has been fixed, but the `0.4` version is still unreleased.

PR #45, opened 2019-09-18
Allow for listening on arbitrary IP address

milleniumbug:
Currently Rust reference library only allows for listening on 0.0.0.0 (Any). PactNet allows for listening on 0.0.0.0 (Any) and 127.0.0.1 (Localhost), but there's no technical reason to not allow listening on specific interface, and possibly on IPv6 addresses too. So I extended for listening on arbitrary IP addresses.


uglyog:
Good one, thanks for the PR!

PR #49, opened 2019-10-06
Fix pack_mock_server C api and usage

jdemeule:
This PR remove the "_ffi" suffix from `create_mock_server` that seems to be readd.
This PR also fix broken usage (javascript and c examples).


uglyog:
Thanks for the PR

PR #53, opened 2019-11-17
Update to Rust 2018 edition

### LINKED ISSUE #52: Upgrade to 2018 edition
The Rust codebase should be upgraded to the Rust 2018 edition, to enable more ergonomic import styles.

audunhalland:
The Rust 2018 edition removes the need to specify `extern crate`, and supports scoped macros, which I tend to like, especially for log statements: `log::info!(...)`. Also, the macro imports can be declared closer to the code that uses the macros (instead of a `macro_use` at the crate root), increasing code readability.

Some upgrades here and there (`log` needs to be at version 0.4 for scoped macro usage).

Fixes #52


audunhalland:
I wanted to have this in place before I start experimenting with async/await support (both internally in `pact_mock_server` and in the public `pact_consumer` API).

uglyog:
LGTM

PR #55, opened 2019-12-17
Async/await

### LINKED ISSUE #51: TODO: Upgrade to std::future
With the whole Rust ecosystem moving over to `std::future` and `async/await`, the pact_consumer library will be considered outdated with the method I added to its public API in `trait StartMockServer`:

```
fn create_mock_server<F>(&self, future_consumer: F) -> ValidatingMockServer
        where F: FnOnce(Box<dyn futures::Future<Item = (), Error = ()> + 'static + Send>);
```
The future is a futures 0.1 future, but should instead be upgraded to a `std::future`.

I think that the pact consumer API should work flawlessly in async code, and I also think this particular API can be designed better. Spawning inside a closure like this appears a little awkward to me. Instead I would like to see the server future being _returned_ instead, something like:

```
let (pact, mock_server_future) = PactBuilder::new(...)
   .interaction(...)
   .create_async();

runtime.spawn(mock_server_future);
```

What's more, the async Rust ecosystem is more than just `tokio`, with `async-std` very recently released as 1.0. Consumer/client code may use async http client based on e.g. `surf` and `async-std`, with no tokio part of the picture at all. In that case I'm not sure that spawning a `hyper` future would work, because hyper expects _tokio_ to be its runtime. I want to think further about possible solutions to this, being an async user of `pact_consumer` myself, and being an advocate of both Pact and Rust.

The final solution can be iterated towards in smaller steps. I can take a look at upgrading `pact_mock_server` to async/await and tokio 0.2 (still in alpha). Then release a new API after tokio 0.2 is out, and then start experimenting with non-tokio runtimes and see what kind of problems arises.

audunhalland:
~~Note that this PR should await a new major stable `reqwest` release before being merged. It should not be long. But I'm creating the PR anyway so that the essential parts can be reviewed.~~

The PR is now ready.

Lib upgrades:
* Requires Rust 1.39.0 for async/await syntax
* Upgrades tokio to 0.2
* Upgrades hyper to 0.13
* Upgrades reqwest to 0.10

API breaking changes:
~~`pact_consumer`: `trait StartMockServer` + `trait ValidatingMockServer`.~~

~~I've tried my best to make consumer tests work nicely with the recently released `tokio-0.2`. As the mock server itself is implemented asynchronously, I feel that it's correct to publicly expose its async interface, and make it compatible with the `[tokio::test]` macro. An publicly asynchronous mock server needs to be shut down explicitly, which I tried to make clear in the docs.~~

Removed `StartMockServer::create_mock_server`

There is no longer any support for starting the `pact_consumer` mock server asynchronously. This idea of mine was flawed. It's now instead always run in a separate thread. Implementation inspired by https://github.com/seanmonstar/reqwest/blob/master/tests/support/server.rs.

This is good because it makes that API completely tokio-independent (no `async` functions exposed).

Fixes #51


rust/pact_verifier/src/pact_broker.rs
```diff
@@ -408,37 +421,44 @@ pub fn fetch_pacts_from_broker(
...
-                })
-                .collect()
+        .map_ok(|(pact_link, pact_json)| {
+            let href = pact_link.href.unwrap();
+            let pact = Pact::from_json(&href, &pact_json);
+            let links = links_from_json(&pact_json);
+
+            (pact, links)
         })
+        .into_stream()
```
audunhalland: Can very easily refactor download of several pact files to be concurrent instead of sequential, if there's any value to that.

audunhalland:
A test seems to be failing on the windows build. I think that test tries to verify that a mock server no longer accepts connections right after it is killed. But I don't know a lot about how the windows platform works..

audunhalland:
It seems I fixed that test fail with the latest commit 🎉

rust/pact_consumer/src/mock_server.rs
```diff
@@ -178,6 +140,148 @@ impl ValidatingMockServer {
...
+        }
+    }
+
+    /// Asynchronously shut down the server.
+    /// This is necessary before the test runtime gets destroyed.
+    /// The reason is that the mock server is spawned as a background task.
+    /// When the tokio test runtime tries to shut down, it will wait for
+    /// all its running tasks to complete. Only by first awaiting the server's
+    /// JoinHandle can the shutdown procedure be properly executed.
+    pub async fn shutdown(mut self) {
```
audunhalland: It appears that this design is flawed after all. While the server shuts down cleanly on test success, it panics when a test panics. That's a panic while panicing, a very bad idea. I think I'll remove the async functionality again, instead focusing on how to reliably run the server in a separate thread. That's the strategy of `reqwest`s own integration tests, so I'll try to copy that behaviour.

uglyog:
Nice, big change. The async code looks a lot cleaner!

uglyog:
@audunhalland I've been having some problems with the aysnc/await changes. With the verifier, the verifier function gets executed, but when it gets to the point to make a HTTP request with Reqwest, it panics with "No current reactor".

In one case, I was able to able to get it to work by then scheduling the Reqwest future with the Tokio task:spawn. That way it gets executed on the current runtime. But with the HalClient, it always panics and I can't get the cause of the panic (it's private in Tokio JoinError).

Switching to the blocking Reqwest client works. Not sure if this is something I'm doing wrong, but I would like to get your thoughts on this issue.

uglyog:
Oh, the one I got working was using Hyper Client directly, not Reqwest.

audunhalland:
Hmm, I will take a look.

audunhalland:
I pushed a new change to the same branch, try now.

PR #56, opened 2020-01-20
Make pact_verifier_cli actually runnable by using tokio::main

audunhalland:
I [commented](https://github.com/pact-foundation/pact-reference/pull/55#issuecomment-575983245) on the `pact_verifier_cli` problem you found. You're right, I had not actually tested running that executable. Here's what I believe is the fix. I have no deep explanation of how the fix works, it's just the first thing I thought of, and it seems to work.


uglyog:
Thanks for the PR

PR #61, opened 2020-03-06
Fix v3 Message Provider States

alilleybrinker:
The v3 spec moves from `providerState` to `providerStates`, but the `Message` implementation in the Rust reference uses the old version. This fixes that, fixes broken documentation links, updates tests to the corrected API, and attempts to at least partially correct the broken auto-derived `Serialize` and `Deserialize` impls (see commit description on that commit for more info).

## Notice

(c) 2020 The MITRE Corporation. All Rights Reserved.

### DIFF
--- a/rust/pact_matching/src/models/matchingrules.rs
+++ b/rust/pact_matching/src/models/matchingrules.rs
@@ -416,6 +416,7 @@ impl Hash for Category {
 
 /// Data structure for representing a collection of matchers
 #[derive(Serialize, Deserialize, PartialEq, Debug, Clone, Eq)]
+#[serde(transparent)]
 pub struct MatchingRules {
     /// Categories of matching rules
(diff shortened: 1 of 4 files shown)


uglyog:
Thanks for the PR!

I didn't use the auto-generated serde serialisation as I wanted full control over the process, but it was more for me than anything wrong with it. I'm happy to move to serde annotations if they work.

alilleybrinker:
Thanks! As I see it, there are three options:

- Serde derives + annotations
- `from_json` functions
- Custom Serde `Serialize` and `Deserialize` impls.

I don't have strong feelings about which way is the best. I do think though it's probably best to pick one and stick with it. Right now the code uses `from_json` consistently, but also does derived `impl`s in places where they don't always match the spec. That's what I was trying to at least partially correct here.

PR #71, opened 2020-08-10
wip: add function to retrieve the CA cert for trusting

mefellows:
This is to allow trusting of the TLS certificate from clients:

1. Adds a new `get_tls_ca_certificate` FFI function
2. Adds a new set of self-signed certificates, alongside the CA (root certificate) that generated them

For example, in Go, I pass in a server config to the test function, that allows the test to configure the HTTP client with the dynamic port as well as get a handle to a pre-configured TLS setup:

```go
	var test = func(config v3.MockServerConfig) error {
		client := &http.Client{
			Transport: &http.Transport{
				TLSClientConfig: config.TLSConfig,
			},
		}
                // make the actual call
		// client.Do(...)
        }
```


mefellows:
Hold off on the merge for a bit, I may need to regenerate a self-signed certificate with a few IP sans.

mefellows:
Actually I think it's fine. One option is to add `127.0.0.1` and other common IP addresses to the SAN list. But wondering if it's better to just ensure `localhost` is set as the hostname in the client calls rather than guess what people will do?

As a side note, `localhost` doesn't appear to be a valid address for the mock server creation call - it fails with invalid address (`-5`). Is that expected?

PR #74, opened 2020-09-10
Add MessagePact (Pact with Messages instead of Interactions)

cstepanian:
This adds an initial implementation of a Message Pact from the v3 spec.
MITRE plans to use this in the near future via FFI to support a basic Pact provider testing capability for message-based services written in C++.

This implementation is missing `merge()`.
Since `merge()` is missing, no method is provided to write a MessagePact to disk and attempt to merge it.
Instead, `overwrite_pact()` is included, which just overwrites a file if it is present.
An additional `merge_pact()` method could be added later.

## Notice

(c) 2020 The MITRE Corporation. All Rights Reserved.

### DIFF
--- a/rust/pact_matching/src/models/message_pact.rs
+++ b/rust/pact_matching/src/models/message_pact.rs
@@ -1,0 +1,359 @@
+//! The `message_pact` module defines a Pact
+//! that contains Messages instead of Interactions.
+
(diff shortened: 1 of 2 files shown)


uglyog:
Ah, the missing bit! Thanks!

PR #76, opened 2020-10-01
Set content-type header in message request to provider

dallagi:
This is a minor change that sets the `Content-Type` http header to `application-json` when sending POST requests to the provider to run the verification process against its messages.

The lack of that header caused some (albeit minor) annoyances in one of our projects, as the library we used to expose the verification endpoint would not automatically parse the body of the request when the header is missing.

I found no tests relevant to this behavior, therefore I only tested this change manually. Please let me know if I missed anything.

Also, I haven't found any guidelines for contributing, so in case I did anything wrong please let me know :)
Thanks!


uglyog (approved):

uglyog:
LGTM

PR #79, opened 2020-11-08
WIP: Feat/pacts for verification

mefellows:
This PR adds the following new features:

1. Adds a new Pact Source to enable pact verification using the new "pacts for verification" endpoint (https://github.com/pact-foundation/pact_broker/issues/307)
1. Pending pacts, WIP pacts and tags are now supported
1. Updates the provider CLI with new these (and other) options (currently does not allow users to specific "selectors" via the CLI, this would be easily achieved after this PR is sorted), which uses the new source
1. Adds basic environment variable support to the verifier to match the existing one

**General comments:**
* The code is still WIP and there are TODOs/Cleanups that need to happen (including moving to that streaming API where it makes sense). It does work though 👍 
* I'd specifically created a new `PactSource` because I wasn't aware of any backwards compatibility issues by changing it. If we don't need that because the only public interface is the CLI, then I'll replace the old implementation instead
* To pass the pending/verification notices to the actual verifier, I've threaded an optional `PactVerificationContext` which contains the notices alongside the Pacts and Pact source through the same process. As there are a few layers in which this passes, it's had an unfortunately side-effect of requiring a lot of functions to change. This probably also needs to be considered for backwards compatibility 
* There are a few places where duplication started appearing as a result of this (e.g. the CLI Main), I'll clean that up

If you have better/other ideas that I could try here, let me know (I considered a number of options, but they didn't feel right).

**Areas I'd like some specific feedback/assistance on:**

* Use of serde_json to serialise/deserialise the payloads. Dealing with the raw `serde_json::Value` was making this code very unreadable. Moving to to `serde_json::to_string` (and `from_value`) made this a lot nicer. Is there a reason this approach wasn't taken elsewhere?
* Where to put things. The pact_broker.rs file is getting pretty big. Currently, I've just dumped some types in there along side the new implementation to fetch pacts. Let me know if you'd like it moved
* Test approach: should we create a Pact test for this? :) Any other pointers? I've tried following previous 
* Code formatting: my editor badly wants to try and reformat the code, but doing so would have made this PR unreadable. How do you do this now?


rust/pact_verifier/src/pact_broker.rs
```diff
@@ -497,11 +508,140 @@ pub async fn fetch_pacts_from_broker(
...
+      consumer_version_selectors: consumer_version_selectors,
+      include_pending_status: pending,
+    };
+    let request_body = serde_json::to_string(&pacts_for_verification).unwrap();
+
+    // Post the verification request
+    let response = match hal_client.find_link("self") {
+      Ok(link) => {
+        // TODO: make this an async stream to make error handling less shit
+        match hal_client.clone().post_json(hal_client.clone().parse_link_url(&link, &hashmap!{})?, request_body).await {
```
uglyog, line 567: If we split this into two statements and introduce a variable for the link URL, then we could get away without all the clones of the hal_client

rust/pact_verifier/src/pact_broker.rs
```diff
@@ -497,11 +508,140 @@ pub async fn fetch_pacts_from_broker(
...
+                  &hal_client.url,
+                  pact_link
+                )
+              )
+            )
+          }
+        })
+        .and_then(|(hal_client, pact_link, context)| async {
+          let pact_json = hal_client.fetch_url(
+            &pact_link.clone(),
```
uglyog, line 627: Because the fetch_url takes references, you might be able to not need the clones with a reference, but just add a ref to the parameters to the closure. But I say might, because the dependency checker is mean.

rust/pact_verifier_cli/src/args.rs
```diff
@@ -121,23 +122,26 @@ pub(crate) fn setup_app<'a, 'b>(program: String, version: &'b str) -> App<'a, 'b
       .help("Consumer name to filter the pacts to be verified (can be repeated)"))
     .arg(Arg::with_name("user")
       .long("user")
+      .env("PACT_BROKER_USERNAME")
```
uglyog, line 125: Oh, wow! That's awesome.

rust/pact_verifier_cli/src/main.rs
```diff
@@ -268,9 +268,30 @@ fn pact_source(matches: &ArgMatches) -> Vec<PactSource> {
           let auth = matches.value_of("user").map(|user| {
             HttpAuth::User(user.to_string(), matches.value_of("password").map(|p| p.to_string()))
           });
-          PactSource::BrokerUrl(name, s!(v), auth, vec![])
+          let pending = matches.is_present("enable-pending");
+          let wip = match matches.value_of("include-wip-pacts-since") {
```
uglyog, line 272: You can just use the map function to change the inner value instead of a pattern match.

rust/pact_verifier_cli/src/main.rs
```diff
@@ -268,9 +268,30 @@ fn pact_source(matches: &ArgMatches) -> Vec<PactSource> {
...
+          };
+          let consumer_version_tags = matches.values_of("consumer-version-tags")
+            .map_or_else(|| vec![], |tags| consumer_tags_to_selectors(tags.map(|tag| tag).collect()));
+          let provider_tags = matches.values_of("provider-tags")
+            .map_or_else(|| vec![], |tags| tags.map(|tag| tag.to_string()).collect());
+
+          PactSource::BrokerWithDynamicConfiguration(name, s!(v), pending, wip, provider_tags, consumer_version_tags, auth, vec![])
         } else if matches.is_present("token") {
-          PactSource::BrokerUrl(s!(matches.value_of("provider-name").unwrap()), s!(v),
+          let name = matches.value_of("provider-name").unwrap().to_string();
```
uglyog, line 283: Should just double check this unwrap won't panic. Is provider-name always present? I can't see from the diff.

rust/pact_verifier_cli/src/main.rs
```diff
@@ -268,9 +268,30 @@ fn pact_source(matches: &ArgMatches) -> Vec<PactSource> {
...
+          let pending = matches.is_present("enable-pending");
+          let wip = match matches.value_of("include-wip-pacts-since") {
```
uglyog, line 285: Use map instead of pattern match

rust/pact_verifier_cli/src/main.rs
```diff
@@ -268,9 +268,30 @@ fn pact_source(matches: &ArgMatches) -> Vec<PactSource> {
...
+            Some(wip) => Some(wip.to_string()),
+            None => None,
+          };
+          let consumer_version_tags = matches.values_of("consumer-version-tags")
+          .map_or_else(|| vec![], |tags| consumer_tags_to_selectors(tags.map(|tag| tag).collect()));
```
uglyog, line 290: `tags.map(|tag| tag).collect()` is mapping but not changing the items, is this just to get a vector? There must be a better way.

uglyog (approved):

rust/pact_verifier/src/lib.rs
```diff
@@ -48,7 +49,9 @@ pub enum PactSource {
     /// Load the pact from a URL
     URL(String, Option<HttpAuth>),
     /// Load all pacts with the provider name from the pact broker url
-    BrokerUrl(String, String, Option<HttpAuth>, Vec<Link>)
+    BrokerUrl(String, String, Option<HttpAuth>, Vec<Link>),
+    /// Load pacts with the newer pacts for verification
+    BrokerWithDynamicConfiguration(String, String, bool, Option<String>, Vec<String>, Vec<ConsumerVersionSelector>, Option<HttpAuth>, Vec<Link>)
```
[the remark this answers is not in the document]
mefellows: Agree, I followed the other pattern but a struct would be nicer.

rust/pact_verifier_cli/src/main.rs
mefellows: I'll double check. If not, I'll ensure the constraint is there in the args, or return an error. It's definitely required in any case.

rust/pact_verifier_cli/src/main.rs
mefellows: 🤦  almost certainly.

uglyog:
> Areas I'd like some specific feedback/assistance on:
>
>    Use of serde_json to serialise/deserialise the payloads. Dealing with the raw serde_json::Value was making this code very unreadable. Moving to to serde_json::to_string (and from_value) made this a lot nicer. Is there a reason this approach wasn't taken elsewhere?

This is more historical than anything else. When the project was started, serde didn't exist and it's predecessor wasn't that great. Serde is awesome, I have not trouble with using it to do the persistence if there is a 1-1 mapping. 

>    Where to put things. The pact_broker.rs file is getting pretty big. Currently, I've just dumped some types in there along side the new implementation to fetch pacts. Let me know if you'd like it moved

I'm pretty bad at that, just keep adding to the same file. I do it in the Kotlin code too, and have been splitting things out when I notice. So, what I'm saying, it won't bother me.

>    Test approach: should we create a Pact test for this? :) Any other pointers? I've tried following previous

Yes! Tests are good. If you look at pact_broker.rs around line 800 you'll find some Pact tests using the Rust consumer DSL. Might be a good idea to break these out into a separate file.

>    Code formatting: my editor badly wants to try and reformat the code, but doing so would have made this PR unreadable. How do you do this now?

I don't. I detest code formatting for no other reason than satisfying someone's OCD. I correct the formatting only for the lines I'm changing to preserve the change history for the file.

mefellows:
Awesome, thanks for the feedback Ron. I'll aim to get stuck into the fixes over the next few days.

rust/pact_verifier/src/pact_broker.rs
mefellows: It turns out the checker is rather mean.

mefellows:
I also added a (very) minor change, to obfuscate the Auth details if put into a print statement:

Screen Shot 2020-11-10 at 1 54 24 pm

uglyog:
Green build, good to merge

PR #80, opened 2020-11-30
feat: return version of the mock server via FFI

mefellows:
Adds a `*char version()` c interface to the mock server FFI, for use with detecting compatible client versions.

NOTE: I originally added it as a `*const  c_char` but it then required freeing via the usual c call. I think this consistency is nicer.

### DIFF
--- a/rust/pact_mock_server_ffi/src/lib.rs
+++ b/rust/pact_mock_server_ffi/src/lib.rs
@@ -74,20 +74,12 @@ use crate::handles::InteractionPart;
 pub mod handles;
 pub mod bodies;
 
-const VERSION: &'static str = env!("CARGO_PKG_VERSION");
+const VERSION: &str = concat!(env!("CARGO_PKG_VERSION"), "\0");
 
 /// Get the current library version
-///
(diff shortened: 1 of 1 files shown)


uglyog:
As it is a compile time static string, I think we can get away without the heap allocation, as long as we can add a NULL terminator to the end. A quick Google found this:

```
pub static VERSION: &str = concat!(env!("VERGEN_SEMVER"), "\0");

#[no_mangle]
pub extern "C" fn get_version() -> *const c_char {
    VER.as_ptr() as *const c_char
}
```

The first line creates a static C string, which can just be returned.

mefellows:
OK cool, that was actually one of the options I had originally. 

The internet seemed to scoff at the manual use of the null terminator, but it didn't bother me. I'll update

uglyog:
I don't understand the scoffing, but if the compiler can create a static string in memory that will never change, why not just return that? We are not going to store it for long, so it should be safe.

mefellows:
Thanks Ron. Sorry, I actually moved this into another PR that I was supposed to raise a few weeks back.

PR #83, opened 2020-12-30
WIP: Feat/add-verifier-ffi

mefellows:
Adds FFI bindings for the verification process. The CLI is great, but especially now that Mac is signing binaries, this approach might simplify things (although, that restriction may also be applied to libs 🤷‍♂️).

I wanted to get your feedback on the approach:

1. It's hijacked the CLI package, which feels wrong, but it allowed a lot of code re-use as you will see. An alternative (I think?) would be to export the functions from another package and have a separate CLI and FFI packages
1. The FFI interface itself simply returns a numeric return code. Preferably, it would actually return a list of interactions that were verified and the related context (e.g. consumer). This would mean the caller could be more granular on feedback (e.g. the current Golang package prints out a test result for each individual interaction, which is useful at a glance).

TODO:
Setup any release step to publish the release binaries to github

_NOTE_: also there were a few minor changes / experiments relating to the above I had locally on my other Mac which died. I think it's backed up to TimeMachine, and will share if it is useful.


rust/pact_verifier_cli/build.rs
```diff
@@ -0,0 +1,41 @@
+use std::env;
```
mefellows, line 1: Note that this file was completely Cargo culted from the other package. I think the crate name of `pact-verifier-ffi` is probably better, but thought there might be a convention to follow.

uglyog:
I was imagining an API where you get a handle to the actual verifier, and control it through that. This is way simpler, almost elegant.

Having a separate FFI crate kind of makes sense, but then we need another one for the shared code, I'm ok with having it in the same one.

PR #84, opened 2020-12-31
feat: allow https scheme and ability to disable ssl verification

mefellows:
Required for https://github.com/pact-foundation/pact-js/issues/561.

I also noticed there was no way to specify a `https` scheme for the CLI, so have added that also.


rust/pact_verifier/src/lib.rs
```diff
@@ -482,7 +487,8 @@ impl <F: RequestFilterExecutor> Default for VerificationOptions<F> {
       provider_version: None,
       build_url: None,
       request_filter: None,
-      provider_tags: vec![]
+      provider_tags: vec![],
+      disable_ssl_verification: None
```
uglyog, line 491: Can just use a boolean defaulted to false instead of an option wrapping a boolean

uglyog (changes requested):
Build has failed

rust/pact_verifier/src/lib.rs
mefellows: Yeah, I dunno what I was thinking to be honest. I think in my head it was briefly that an optional type would mean no changes to the interface (i.e. backwards compatible) but of course that's just stupid. Updated.

PR #85, opened 2021-01-03
Multipart lib upgrade

audunhalland:
Tokio has already reached 1.0, so in my own rust app, that's using pact-reference, I started to prepare for this, by first analyzing outdated dependencies (hoping to improve debug build time).

pact-reference stood out, as it pulls in tokio 0.1 and older versions of hyper. This is caused by the lib `formdata`, that depends on a very old version of hyper, and it seems that the lib itself is sort of unmaintained.

I've replaced the lib with the slightly more low-level https://docs.rs/multipart, which can be compiled without hyper support.

Note breaking API change in `pact_mock_server_ffi`.


uglyog:
Awesome! I once tried to upgrade that crate to use the http crate instead of hyper, but gave up. It's good to be able to remove it.

I have a chore/upgrade-tokio branch with token + hyper upgraded to the latest, but I'm having some trouble because we're using rustls instead of native-tls, and the hyper support for it is not great.

audunhalland:
Cool that you already started with tokio 1. I'm really eager to see all of these APIs stabilized. You can let me know if there's any way I can help, though my rustls knowledge is 0.

pact-reference is a great way for me to contribute to the rust community, every "hot crate" out there has devs racing to be the one to make those crucial PRs, I'm too slow. This project is a bit more quiet :)

PR #86, opened 2021-01-07
fix: request headers generated unexpected differences 

stevecooperorg:
Request headers are serialised in a hashmap; serializing and deserializing rust hashmaps can change the internal order of keys in the hashmap. 

This can confuse the Request::differences_from code, which sees out-of-order keys as different. 

Sorting the keys before comparison fixes the problem.


uglyog:
Thanks for the PR

PR #89, opened 2021-01-09
chore: get rid of nom dupes by using tree_magic_mini

audunhalland:


audunhalland:
I'll combine different upgrades into 1 PR instead.

PR #93, opened 2021-02-24
Set install_name on osx lib during cargo build

mefellows:


uglyog:
It will need to be wrapped in an if to only run for OSX, can use the os_info crate for this: https://crates.io/crates/os_info

mefellows:
Yep, it would. FWIW in pact-go, I'm manually setting the install path prior to use on OSX: https://github.com/pact-foundation/pact-go/blob/feat/v3-ffi-verifier/v3/installer/installer.go#L209-L220.

I'll need to do more digging, because it would be ideal if I can remove the hard coding altogether so other client libs don't need to do it. I tried doing that, but couldn't make it work with the spike time I had.

uglyog:
I guess with OSX apps this happens on install

mefellows:
@uglyog I think I have a working solution now. After setting the install name, the library can be dropped into a known library search path on a Mac (see https://opensource.apple.com/source/dyld/dyld-519.2.1/src/dyld.cpp.auto.html) or if desired by the client library, installed into a custom location without it complaining about absolute references.

mefellows:
On a related note, do you know where it would be possible to set the `-current_version` flag for the dylib (and a similar thing for linux/windows libs?). One of the issues with Go, is that you can't have a package depend on a library run in order to download, install and then check it. There's ways around it, but one option I was exploring was adding version data into the exported lib, and then using native tools (like `otool -L`) to get the current version. Alternatively, we could version the file itself (but that's less than ideal except for major versions).

https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/DynamicLibraries/100-Articles/DynamicLibraryDesignGuidelines.html#//apple_ref/doc/uid/TP40002013-SW23

uglyog:
I know that Windows DLLs suport a version in the DLL metadata, but I don't know about Linux shared objects.

uglyog:
Linux doesn't support anything like that

> Essentially, version information is not (or not primarily, don't know from my head) encoded in the library itself, but rather in its filename. Version numbers are normally given in three-dot format, with the major number increasing for each break in downward ABI compatibility, the middle for breaks in upward ABI compatibility, and the minor for patches that did not change the ABI.

mefellows:
Oh well, thanks Ron. I'll find another way. For future context, the idea was that after downloading the package, I would then invoke the `version` ffi function to ensure the correct one was installed. I can't do it because of the chicken-and-egg linking problem described above. I can separate packages I think to still do it, but not as nicely. 

Anyway, this PR is good to go from my side.

PR #95, opened 2021-03-13
fix: move state params to "params" key in provider state change executor

mefellows:
Given this state from the `providerStates` key in the pact:

i.e. I was expecting something to mirror this:
```
{
  "name": "User foo exists",
  "params": {
    "some": "param"
  }
}

```

The current state change process for FFI / CLI sends the following message to the pact client to instruct it to prepare a state:

```
{ "action": "teardown", "some": "param", "state": "User foo exists" }
```

I encountered two issues with this:

1. Note the parameters aren't self-contained for simple extraction to pass to a user provided function. Not having a clear top-level structure made the JSON marshaller a bit confused, so extra work was required to separate the "framework" part from the "user supplied data"
2. If a user wants to use the keys `state` or `action` we'll have a conflict

I'm proposing that we move the body into a top level `params` key, that frameworks can then parse.

I'm thinking it probably makes sense to have both whilst, but not sure of the adoption here and if it will break other users.

P.S. unrelated, but the related "value from provider states" stuff (just return JSON from the state handler) was a piece of magic and so easy to support👌

### DIFF
--- a/rust/pact_verifier/src/callback_executors.rs
+++ b/rust/pact_verifier/src/callback_executors.rs
@@ -85,20 +85,15 @@ impl ProviderStateExecutor for HttpRequestProviderStateExecutor {
       Some(state_change_url) => {
         let mut state_change_request = Request { method: "POST".to_string(), .. Request::default() };
         if self.state_change_body {
-          let mut json_body = json!({
-                    "state".to_string() : json!(provider_state.name.clone()),
-                    "action".to_string() : json!(if setup {
+          let json_body = json!({
+                    "state".to_string() : provider_state.name.clone(),
+                    "params".to_string() : provider_state.params.clone(),
+                    "action".to_string() : if setup {
                         "setup".to_string()
                     } else {
                         "teardown".to_string()
-                    })
+                    }
                 });
-          {
-            let json_body_mut = json_body.as_object_mut().unwrap();
-            for (k, v) in provider_state.params.clone() {
-              json_body_mut.insert(k, v);
-            }
-          }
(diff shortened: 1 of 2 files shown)


rust/pact_verifier/src/callback_executors.rs
```diff
@@ -85,20 +85,22 @@ impl ProviderStateExecutor for HttpRequestProviderStateExecutor {
...
                         "teardown".to_string()
                     })
                 });
-          {
-            let json_body_mut = json_body.as_object_mut().unwrap();
-            for (k, v) in provider_state.params.clone() {
-              json_body_mut.insert(k, v);
-            }
-          }
+          // leave for backwards compatability?
```
uglyog, line 97: Not many people are using this now, so it might be better to just change it.

rust/pact_verifier/src/callback_executors.rs
```diff
@@ -85,20 +85,22 @@ impl ProviderStateExecutor for HttpRequestProviderStateExecutor {
       Some(state_change_url) => {
         let mut state_change_request = Request { method: "POST".to_string(), .. Request::default() };
         if self.state_change_body {
-          let mut json_body = json!({
+          let json_body = json!({
                     "state".to_string() : json!(provider_state.name.clone()),
+                    "params".to_string() : json!(provider_state.params.clone()),
```
uglyog, line 90: I've discovered that the serde `json!` macro is recursive, so we don't need to repeat it on the values.

mefellows:
Good to go I think?

PR #96, opened 2021-04-07
refactor: Use Anyhow instead of `io::Result`

cstepanian:
- Add a dependency on the `anyhow` library.
- Replace `io::Result` with `anyhow::Result` in functions that can return multiple types of error.
- Refactor `load_pact_from_url` and `load_pact_from_json` to use `?` operator instead of pattern matching.

## Notice


uglyog:
Nice! That's way better. Thank you.

PR #98, opened 2021-05-03
feat: allow ffi bindings to set spec version

mefellows:
This allows for clients to build multi version compatible interfaces via the FFI bindings.

I wasn't so sure about creating a new enum for the `SpecificationVersion` that is exposed to the c interface.

Alternatively, we could expose the underlying `PactSpecification` enum. Let me know what you prefer.

### DIFF
--- a/rust/pact_mock_server_ffi/src/lib.rs
+++ b/rust/pact_mock_server_ffi/src/lib.rs
@@ -69,6 +69,7 @@ use pact_matching::time_utils::{parse_pattern, to_chrono_pattern};
 use pact_mock_server::{MANAGER, MockServerError, tls::TlsConfigBuilder, WritePactFileErr};
 use pact_mock_server::server_manager::ServerManager;
 use pact_models::bodies::OptionalBody;
+use pact_models::PactSpecification;
 
 use crate::bodies::{empty_multipart_body, file_as_multipart_body, MultipartBody, process_json, request_multipart, response_multipart};
 use crate::handles::InteractionPart;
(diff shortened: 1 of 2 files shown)


uglyog:
There is also a `PactSpecification` enum in `pact_matching_ffi`, so that will make 3 of them. Feels like the wrong thing to do, maybe we can just add `pact_matching_ffi` as a dependency.

uglyog:
Or just add the `#[repr(C)]` annotation to the one in the models. I think that just forces the memory layout to conform to C.

uglyog:
Interesting, from the Rust docs:
> Warning: There are crucial differences between an enum in the C language and Rust's field-less enums with this representation. An enum in C is mostly a typedef plus some named constants; in other words, an object of an enum type can hold any integer value. For example, this is often used for bitflags in C. In contrast, Rust’s field-less enums can only legally hold the discrimnant values, everything else is undefined behavior. Therefore, using a field-less enum in FFI to model a C enum is often wrong.

I guess this is because on the other side you can assign any integer value to the enum

mefellows:
I've just made the change. It compiles and works in the go consumer I have, so I think it's 🆗

uglyog:
I'll think I'll change it to use the one from `pact_matching_ffi` so there is no FFI influences in the core Rust code.

TimothyJones:
> so there is no FFI influences in the core Rust code.

This feels right. I think conceptually, there's a translation layer between the FFI bindings and the core - which would avoid bleeding the subtleties between the two types of enums.

mefellows:
Cool, thanks Ron. I'll leave this with you then? Do you want to merge it in and fix or want me to address?

uglyog:
I can do it, you'll just need to check it works ok

mefellows:
👍

uglyog:
Sorted with a feature flag: 82711d6

PR #100, opened 2021-05-16
Feat/more consumer ffi matchers

mefellows:
Supports matchers to be passed in as JSON to the FFI for:

* Request Path, Query and Headers
* Response Headers
* Adds a basic feature test to show how the FFI matchers may be used

(bodies were already supported)

### DIFF
--- a/rust/pact_matching/src/models/matchingrules.rs
+++ b/rust/pact_matching/src/models/matchingrules.rs
@@ -214,8 +214,6 @@ impl <T: Display> DisplayForMismatch for &[T] {
 /// Set of all matching rules
 #[derive(Serialize, Deserialize, Debug, Clone, Eq)]
 pub enum MatchingRule {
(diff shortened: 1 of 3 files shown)


rust/pact_matching/src/models/matchingrules.rs
```diff
@@ -214,6 +214,8 @@ impl <T: Display> DisplayForMismatch for &[T] {
 /// Set of all matching rules
 #[derive(Serialize, Deserialize, Debug, Clone, Eq)]
 pub enum MatchingRule {
+  /// Dummy rule to allow all values passed in via the FFI to be JSON integration values, even primitives
+  Dummy,
```
uglyog, line 218: There must be a better way. I'll have a look into it.

mefellows: I figured you might say that. Thanks, any ideas would be great.

PR #104, opened 2021-05-26
Feat/consumer version selectors

mefellows:
* Supports version selectors via CLI and FFI interface (JSON blobs, as per Ruby CLI)
* State callback timeout can now be set via flag
* Add env vars for provider request filtering (see https://github.com/pact-foundation/pact-js/pull/668#issuecomment-848475693)


mefellows:
cc: @TimothyJones

uglyog (changes requested):
LGTM

The only thing to change would be the `request-timeout` is not validated, and if rubbish is passed in it will silently default to `5000`.

mefellows:
> LGTM
> 
> The only thing to change would be the `request-timeout` is not validated, and if rubbish is passed in it will silently default to `5000`.

Yes I was wondering. What do you suggest is the best approach here (I modelled it somewhat off the "port" field).

Should I log an error (via `error!`) and return an `Err` result, or something else?

Ideally, the argument validation logic is in one place, but it's kind of spread across `args.rs` and `verifier.rs`. Maybe I move it into `args.rs` and have a field validator there?

uglyog:
It just needs the some validator as per the port, which checks for integers.

mefellows:
Ha, timing. Yes I realised almost immediately after. See b3d4da67

uglyog (approved):

uglyog:
Builds failing:
```
error[E0433]: failed to resolve: use of undeclared crate or module `env`
Error:    --> pact_matching/src/models/message.rs:496:16
    |
496 |     let file = env::current_dir()
    |                ^^^ use of undeclared crate or module `env`
```

mefellows:
Yeah i'll fix that. strange, my local compiler whinges when I add that import that it's an unused import and compiles fine without it 🤷

uglyog:
Also, `message_with_binary_body_serialises.zip` is missing.

Changing that to 
```
let file = Path::new(env!("CARGO_MANIFEST_DIR"))
        .join("tests/data")
        .join("message_with_binary_body_serialises.zip")
        .to_owned();
```
means the tests can also be run from the project root.

uglyog:
`pact_matching/src/models/message_pact.rs:503` fails because it has the Pact-Rust version hard-coded in the expected string

mefellows:
Thanks, leave with me I'll sort that out

PR #105, opened 2021-05-26
WIP: feat: allow ffi consumers to log using core

mefellows:
Interested in feedback here on this idea.

cc: @TimothyJones - let's see how our experiment goes with Pact JS to see how we should/could evolve it. Probably worth marking these methods as highly experimental _just in case_.

I've just added a namespace (target in the `log` package) to the log, to be able to differentiate the client logs from the core from the library. We might need to come up with patterns here, but let's see how this goes.

from https://docs.rs/log/0.4.14/log/:
> A log request consists of a target, a level, and a body. A target is a string which defaults to the module path of the location of the log request, though that default may be overridden. Logger implementations typically use the target to filter requests based on some user configuration.

### DIFF
--- a/rust/pact_mock_server_ffi/src/lib.rs
+++ b/rust/pact_mock_server_ffi/src/lib.rs
@@ -130,15 +130,19 @@ pub unsafe extern fn log_level(level: *const c_char) {
 
 /// Log using the shared core logging facility, instead
 ///
+/// * `source` - String. The source of the log, such as the class or caller framework to
+///                      disambiguate log lines from the rust logging (e.g. pact_go)
 /// * `log_level` - String. One of TRACE, DEBUG, INFO, WARN, ERROR
 /// * `message` - Message to log
 ///
 /// Exported functions are inherently unsafe.
 #[no_mangle]
-pub unsafe extern fn log(log_level: *const c_char, message: *const c_char) {
+pub unsafe extern fn log(source: *const c_char, log_level: *const c_char, message: *const c_char) {
+  let target = convert_cstr("target", source).unwrap_or("client");
+
   if !message.is_null() {
     match convert_cstr("message", message) {
-      Some(message) => log!(log_level_from_c_char(log_level), "{}", message),
+      Some(message) => log!(target: target, log_level_from_c_char(log_level), "{}", message),
       None => ()
     }
   }
--- a/rust/pact_mock_server_ffi/src/lib.rs
+++ b/rust/pact_mock_server_ffi/src/lib.rs
@@ -56,8 +56,7 @@ use std::ffi::CString;
 use std::panic::catch_unwind;
 use std::ptr::null_mut;
 use std::str;
-use std::path::PathBuf;
-
+use std::str::FromStr;
 use bytes::Bytes;
 use chrono::Local;
 use env_logger::Builder;
@@ -102,13 +101,9 @@ pub extern "C" fn version() -> *const c_char {
 #[no_mangle]
 pub unsafe extern fn init(log_env_var: *const c_char) {
   let log_env_var = if !log_env_var.is_null() {
-    let c_str = CStr::from_ptr(log_env_var);
-    match c_str.to_str() {
-      Ok(str) => str,
-      Err(err) => {
-        warn!("Failed to parse the environment variable name as a UTF-8 string: {}", err);
-        "LOG_LEVEL"
-      }
+    match convert_cstr("log env var", log_env_var) {
+      Some(level) => level,
+      None => "LOG_LEVEL"
     }
   } else {
     "LOG_LEVEL"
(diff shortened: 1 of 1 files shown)


rust/pact_mock_server_ffi/src/lib.rs
```diff
@@ -111,6 +108,56 @@ pub unsafe extern fn init(log_env_var: *const c_char) {
   builder.try_init().unwrap_or(());
 }
 
+/// Sets the log level explicitly
+///
+/// # Safety
+///
+/// Exported functions are inherently unsafe.
+#[no_mangle]
+pub unsafe extern fn log_level(level: *const c_char) {
```
uglyog, line 117: Maybe re-name this, as it initialises the logging, which can only be done once. I.e., someone might think they can call it again later to change the log level, and that won't do anything.

uglyog (approved):

mefellows:
ah yes, good call. I'll have a play with the ffi matching stuff to see how it could all work nicely together. This should have been marked as a draft PR but I don't think you can change after the fact?

rust/pact_mock_server_ffi/src/lib.rs
```diff
@@ -111,6 +108,56 @@ pub unsafe extern fn init(log_env_var: *const c_char) {
...
+/// Log using the shared core logging facility, instead
+///
+/// * `source` - String. The source of the log, such as the class or caller framework to
+///                      disambiguate log lines from the rust logging (e.g. pact_go)
+/// * `log_level` - String. One of TRACE, DEBUG, INFO, WARN, ERROR
+/// * `message` - Message to log
+///
+/// Exported functions are inherently unsafe.
+#[no_mangle]
+pub unsafe extern fn log(source: *const c_char, log_level: *const c_char, message: *const c_char) {
```
adamrodger, line 134: I think having to supply the source from the parent library is a bit awkward on every line. Perhaps you'd end up wrapping it in the parent library or something so it always returns the same thing? It's kinda unclear what a sensible value would be to pass in though.

Is there any way we can make the log level an `enum` as well? It seems a bit brittle that you could pass it incorrectly and then it just defaults to INFO.

rust/pact_mock_server_ffi/src/lib.rs
```diff
@@ -111,6 +108,56 @@ pub unsafe extern fn init(log_env_var: *const c_char) {
...
+    match convert_cstr("message", message) {
+      Some(message) => log!(target: target, log_level_from_c_char(log_level), "{}", message),
+      None => ()
+    }
+  }
+}
+
+unsafe fn log_level_from_c_char(log_level: *const c_char) -> log::Level {
+  if !log_level.is_null() {
+    let level = match convert_cstr("log_level", log_level) {
```
adamrodger, line 147: You can use `unwrap_or_default` in this call and the one beneath instead of doing the pattern match. There should be a lint for that I think, but dunno if this crate has customised the lint rules?

rust/pact_mock_server_ffi/src/lib.rs
mefellows: I can look at making it an enum. Would it actually make it safer, my understand is that the c caller could still just pass in any integer anyway? I haven't worked much with c interfaces, so if it's more idiomatic happy to change.

As for the source. The problem at the moment, is that the Rust log file will log the package the message relates to which is helpful. If we logged from a client without passing it through, we lose the context. So a better name could be "package", but I would imagine sending a value that indicates where the source of the log came from.

e.g. given a log file that looks like this

```
[2021-05-27T10:51:33Z TRACE pact_matching::models::matchingrules] matcher_is_defined: for category header and path ["$", "Content-Type"] -> false
[2021-05-27T10:51:33Z TRACE pact_matching::models::matchingrules] Calculating weight for path tokens '[Root, Field("Authorization")]' and path '["$", "Authorization"]'
[2021-05-27T10:51:33Z TRACE pact_matching::models::matchingrules] Calculated weight (4, 2) for path 'Authorization' and '["$", "Authorization"]'
[2021-05-27T10:51:33Z TRACE pact_matching::models::matchingrules] matcher_is_defined: for category header and path ["$", "Authorization"] -> true
[2021-05-27T10:51:33Z TRACE pact_matching::models::matchingrules] Calculating weight for path tokens '[Root, Field("Authorization")]' and path '["$", "Authorization"]'
[2021-05-27T10:51:33Z TRACE pact_matching::models::matchingrules] Calculated weight (4, 2) for path 'Authorization' and '["$", "Authorization"]'
[2021-05-27T10:51:33Z DEBUG pact_matching::matchers] String -> String: comparing 'Bearer 1234' to 'Bearer 1234' using Type
[2021-05-27T10:51:33Z DEBUG pact_matching] --> Mismatches: []
[2021-05-27T10:51:33Z DEBUG pact_mock_server::hyper_server] Test context = {"mockServer": Object({"href": String("https://127.0.0.1:8080"), "port": Number(8080)})}
[2021-05-27T10:51:33Z INFO  pact_mock_server::hyper_server] Request matched, sending response Response ( status: 200, headers: Some({"Content-Type": ["application/json"]}), body: Present(300 bytes) )
[2021-05-27T10:51:33Z DEBUG pact_mock_server::hyper_server]      body: '{"accountBalance":123.76,"arrayContaining":["string",1,{"foo":"bar"}],"dateTime":"2020-01-01","equality":"a thing","id":12,"itemsMin":["thereshouldbe3ofthese","thereshouldbe3ofthese","thereshouldbe3ofthese"],"itemsMinMax":[27,27,27,27,27],"lastName":"LastName","name":"FirstName","superstring":"foo"}'
[2021-06-27T10:51:123 DEBUG client::pact_go::somepackage::nestedpackage] a debug message
[2021-06-27T10:51:123 DEBUG client::pact_go::somepackage::nestedpackage] passinsg argumetns {...} to FFI
[2021-05-27T10:51:33Z INFO  pact_mock_server::mock_server] Writing pact out to '/Users/matthewfellows/go/src/github.com/pact-foundation/pact-go/examples/v3/pacts/V3Consumer-V3Provider.json'
[2021-05-27T10:51:33Z DEBUG pact_matching::models] Writing new pact file to "/Users/matthewfellows/go/src/github.com/pact-foundation/pact-go/examples/v3/pacts/V3Consumer-V3Provider.json"
[2021-06-27T10:51:123 DEBUG client::pact_go::somepackage::nestedpackage] mock server cleaning up port: 8080
[2021-05-27T10:51:33Z DEBUG pact_mock_server::server_manager] Shutting down mock server with port 8080
[2021-05-27T10:51:33Z DEBUG pact_mock_server::server_manager] Shutting down mock server with port 8080 - MockServerMetrics { requests: 1 }
[2021-05-27T10:51:33Z DEBUG pact_mock_server::mock_server] Mock server 54ec5bc8-aaf3-4ed6-ae75-4265609ed51e shutdown - MockServerMetrics { requests: 1 }
[2021-05-27T10:51:33Z TRACE mio::poll] deregistering event source from poller
[2021-05-27T10:51:33Z DEBUG rustls::session] Sending warning alert CloseNotify
[2021-05-27T10:51:33Z TRACE mio::poll] deregistering event source from poller
```

Here you can easily see which packages of the `pact-go` library are logging it out. But perhaps there are other options? There are other options, e.g. to log the source file that was logging itself. I'm not sure what's best.

What do you think?

PR #111, opened 2021-05-31
test: extract_headers function, specially with comma separated values

individual-it:
test the `extract_headers` function. Specially when a comma separated list of values is given, see https://github.com/pact-foundation/pact-js/issues/646

CC @tonynguyenit18


uglyog:
Awesome!

PR #117, opened 2021-06-26
feat: Update PHP example

tienvx:
* Require json extension
* Use c's header files from pact-reference's release
* Separate http and message for consumer and provider names
* Don't override pact files
* Call `with_specification`
* Question for `size` parameter


php/src/consumer-2.php
```diff
@@ -28,7 +30,7 @@
         "regex": "^[0-9a-f]{8}(-[0-9a-f]{4}){3}-[0-9a-f]{12}$",
         "value": "fb5a885f-f7e8-4a50-950f-c1a64a94d500"
     }
-}', 36); // size of uuid
+}', 0); // is size necessary?
```
uglyog, line 33: Size is not needed with JSON, because the body is a C string. With binary bodies, we need to know the number of bytes in the body.

I have updated the comments on the function to describe this

PR #118, opened 2021-07-03
proposal: all logging methods to pact_matching_ffi

mefellows:
I really just want to add a `log_message` function so that callers can utilise the rust logic, but in the course of doing that, it felt unusual that logging was split across packages.

Equally, it doesn't feel write to have these other initialisers in the matching package which seem to have a different flavour of initialisation altogether and I'm a little out of my depth there.

This is a quick PR to test the waters - is this something we should be thinking about?

I'm also wondering if it just makes sense to have a single FFI that contains all of the functions needed for a standard "pact" client?

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1125,6 +1125,7 @@ version = "0.0.5"
 dependencies = [
  "anyhow",
  "bytes",
+ "env_logger",
  "fern",
  "lazy_static",
  "libc",
--- a/rust/pact_matching_ffi/Cargo.toml
+++ b/rust/pact_matching_ffi/Cargo.toml
@@ -23,6 +23,7 @@ log = "0.4.8"
 serde_json = "1.0.51"
 bytes = "1.0.1"
 lazy_static = "1.4.0"
+env_logger = "0.8.2"
 
 [lib]
 crate-type = ["rlib", "cdylib", "staticlib"]
--- a/rust/pact_matching_ffi/src/log/ffi.rs
+++ b/rust/pact_matching_ffi/src/log/ffi.rs
@@ -17,6 +17,9 @@ use crate::log::logger::{add_sink, apply_logger, set_logger};
 use crate::log::sink::Sink;
 use crate::log::status::Status;
 use crate::util::string::to_c;
+use log::*;
+use std::str::FromStr;
+use env_logger::Builder;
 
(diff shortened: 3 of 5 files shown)


uglyog:
That's a good point. A single FFI library makes sense. We should probably do that and merge all the FFI libs into one before we continue too much further.

mefellows:
> That's a good point. A single FFI library makes sense. We should probably do that and merge all the FFI libs into one before we continue too much further.

I think I'll convert this into a separate PR to do the "single FFI" thing.

cc: @yanncourtel @adamrodger @TimothyJones @tienvx 

What are your thoughts on a single FFI (rather than currently 2?). This in theory should create smaller distributable packages and clear the FFI interface (particularly with the overlap in the logging interfaces required by both the consumer / provider side).

TimothyJones:
> What are your thoughts on a single FFI (rather than currently 2?). This in theory should create smaller distributable packages and clear the FFI interface (particularly with the overlap in the logging interfaces required by both the consumer / provider side).

I think this is definitely the best option - it'll optimise the overall space required for the binaries in a pact distribution, and mean that we don't have to consider how to maintain state (eg logging levels) across multiple libs. I guess a downside could be that if you only need one or two pact functions, the overall linked process size is potentially larger?

Also, as I mentioned on the other issue, I think we should namespace the ffi functions with a prefix `pact_`, or potentially `pact_matching_`, `pact_verifier_` etc

tienvx:
If merging libraries (`libpact_mock_server_ffi` and `libpact_verifier_ffi`) bring benefits to maintainers, I think it's okay. For users like me, I think the only benefit right now is less files to download. Currently  we have to download at least 4 files:
* pact_verifier_ffi-c.h
* pact_mock_server_ffi-c.h
* libpact_verifier_ffi-linux-x86_64.so.gz
* libpact_mock_server_ffi-linux-x86_64.so.gz

I expect there is only one file to download:
* ffi-linux-x86_64.tar.gz

I don't know what `pact_matching_ffi` do and I didn't use it so I can't comment on it.

adamrodger:
Definitely one FFI lib is best for PactNet. You can't really maintain
global state like logging levels properly without it, plus it simplifies
the integration glue code

uglyog:
Closing this PR as it is no longer relevant?

PR #119, opened 2021-07-03
feat: allow callers to add messages to log

mefellows:
See also #118.

I went with `log_message` instead of `log` because Golang warned that it conflicted with default libraries on a Mac, so felt it was best to avoid a general term.

### DIFF
--- a/rust/pact_mock_server_ffi/src/lib.rs
+++ b/rust/pact_mock_server_ffi/src/lib.rs
@@ -51,7 +51,7 @@ use std::ffi::CString;
 use std::panic::catch_unwind;
 use std::path::PathBuf;
 use std::ptr::null_mut;
-use std::str::from_utf8;
+use std::str::{from_utf8, FromStr};
 
 use bytes::Bytes;
 use chrono::Local;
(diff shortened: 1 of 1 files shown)


mefellows:
See also #118. I think we should get this in first (if acceptable) because we are using theses different FFI packages already.

We can migrate this to the single FFI package if/when we do that.

TimothyJones:
Would it not make more sense to make the single package before doing further ffi client work?

Re: naming collisions, I think the usual c convention is to include a namespace prefix in the names of everything in the external interface (eg `pact_`).

mefellows:
> Would it not make more sense to make the single package before doing further ffi client work?

I don't think so. Let's not have perfect be the enemy of the good! Lots of little cleanup etc. work to do, refactoring packages, creating crates, release scripts etc.

We already have an established process. This would at least allow everyone to continue and can migrate to a single library when it's ready.

> Re: naming collisions, I think the usual c convention is to include a namespace prefix in the names of everything in the external interface (eg pact_).

That's a good idea for #118.

uglyog (approved):

PR #120, opened 2021-07-05
Support verifying message metadata via verifier (CLI/FFI)

mefellows:
There is currently no way to pass metadata back to the (HTTP based) verifier, for use by the command line or FFI interface.

This PR adds that.

1. I noticed the verifier doesn't read the `metadata` section (still uses `metaData` when reading a Pact) but generates a pact with `metadata`. It now reads both for compatibility
2. The incoming payload now supports a header (`PACT_MESSAGE_METADATA`) that accepts a base64 encoded JSON of key/value pairs. The values may be complex JSON objects, or simple primitives
3. The `metadata` property of the `Message` struct, now supports complex values to support this.

### DIFF
--- a/rust/pact_matching/src/lib.rs
+++ b/rust/pact_matching/src/lib.rs
@@ -354,6 +354,7 @@ use pact_models::content_types::ContentType;
 use pact_models::generators::{apply_generators, GenerateValue, GeneratorCategory, GeneratorTestMode, VariantMatcher};
 use pact_models::matchingrules::{calc_path_weight, Category, MatchingRule, MatchingRuleCategory, path_length, RuleList};
 use pact_models::PactSpecification;
+use pact_models::json_utils::json_to_string;
 
 use crate::headers::{match_header_value, match_headers};
 use crate::matchers::*;
@@ -1422,7 +1423,7 @@ pub fn match_message_metadata(
         },
         None => {
           result.insert(key.clone(), vec![Mismatch::MetadataMismatch { key: key.clone(),
-            expected: value.as_str().unwrap_or_default().to_string(),
+            expected: json_to_string(&value),
             actual: "".to_string(),
             mismatch: format!("Expected message metadata '{}' but was missing", key) }]);
         }
--- a/rust/pact_matching/src/lib.rs
+++ b/rust/pact_matching/src/lib.rs
@@ -1403,13 +1403,13 @@ pub fn match_message_metadata(
     expected.contents.metadata
   } else {
     expected.as_message().unwrap().metadata.iter()
-      .map(|(k, v)| (k.clone(), Value::String(v.clone()))).collect()
+      .map(|(k, v)| (k.clone(), v.clone())).collect()
   };
   let actual_metadata = if let Some(actual) = actual.as_v4_async_message() {
     actual.contents.metadata.clone()
   } else {
     actual.as_message().unwrap().metadata.iter()
-      .map(|(k, v)| (k.clone(), Value::String(v.clone()))).collect()
+      .map(|(k, v)| (k.clone(), v.clone())).collect()
   };
   debug!("Matching message metadata. Expected '{:?}', Actual '{:?}'", expected_metadata, actual_metadata);
 
--- a/rust/pact_matching/src/lib.rs
+++ b/rust/pact_matching/src/lib.rs
@@ -1411,6 +1411,8 @@ pub fn match_message_metadata(
     actual.as_message().unwrap().metadata.iter()
       .map(|(k, v)| (k.clone(), Value::String(v.clone()))).collect()
   };
+  debug!("Matching message metadata. Expected '{:?}', Actual '{:?}'", expected_metadata, actual_metadata);
+
   if !expected_metadata.is_empty() || context.config == DiffConfig::NoUnexpectedKeys {
     for (key, value) in &expected_metadata {
       match actual_metadata.get(key) {
--- a/rust/pact_matching/src/models/message.rs
+++ b/rust/pact_matching/src/models/message.rs
@@ -183,14 +183,14 @@ impl Message {
                     None => format!("Message {}", index)
                 };
                 let provider_states = ProviderState::from_json(json);
-                let metadata = match json.get("metaData") {
-                    Some(&Value::Object(ref v)) => v.iter().map(|(k, v)| {
-                        (k.clone(), match v {
-                            &Value::String(ref s) => s.clone(),
-                            _ => v.to_string()
-                        })
-                    }).collect(),
-                    _ => hashmap!{}
+                let metadata = match json.get("metaData").or(json.get("metadata")) {
+                  Some(&Value::Object(ref v)) => v.iter().map(|(k, v)| {
+                      (k.clone(), match v {
+                          &Value::String(ref s) => s.clone(),
+                          _ => v.to_string()
+                      })
+                  }).collect(),
+                  _ => hashmap!{},
                 };
                 Ok(Message {
                   id: None,
--- a/rust/pact_matching/src/models/message.rs
+++ b/rust/pact_matching/src/models/message.rs
@@ -47,7 +47,7 @@ pub struct Message {
 
     /// Metadata associated with this message.
     #[serde(default)]
-    pub metadata: HashMap<String, String>,
+    pub metadata: HashMap<String, Value>,
 
     /// Matching rules
     #[serde(rename = "matchingRules")]
@@ -125,7 +125,7 @@ impl Interaction for Message {
       contents: MessageContents {
         contents: self.contents.clone(),
         metadata: self.metadata.iter()
-          .map(|(k, v)| (k.clone(), Value::String(v.clone())))
+          .map(|(k, v)| (k.clone(), v.clone()))
           .collect(),
         matching_rules: self.matching_rules.rename("body", "content"),
         generators: self.generators.clone()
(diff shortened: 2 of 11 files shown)


TimothyJones:
> I think we could support that being first trying to unmarshal an object of the type above, and falling back to current behaviour otherwise - thoughts?

I think `contents` and `metadata` are too generic, so if we switch behaviour on those names then we'll do the wrong thing sometimes.

Also, I think the example isn't correct: you have

```
{
  "id": 127,
  "event": "some message",
  "__pactMetadata": {
    "Content-Type": "application/json",
    "topic": "baz"
  }
}
```

but to match the implementation in pact-js, it would be:

```
{
  "message": { 
     "id": 127,
     "event": "some message",
  },
  "__pactMessageMetadata": {
    "Content-Type": "application/json",
    "topic": "baz"
  }
}
```

(although now that I read the PR again, perhaps we're not on the same page about the original proposal? I was proposing the wrapped payload, but with less generic field names).

TimothyJones:
This is where I added the wrapper: https://github.com/pact-foundation/pact-js/commit/4aa4e9faf6f35887b365bcb830e2f8ed9eb8c25f#diff-0b4824334fab898c44e8a3dd93c2c7a0408727ef2a1a1e1ca141de5c3c6cd226R37

rust/pact_matching/src/models/message.rs
```diff
@@ -183,7 +183,15 @@ impl Message {
                             _ => v.to_string()
                         })
                     }).collect(),
-                    _ => hashmap!{}
+                    _ => match json.get("metadata") {
```
uglyog, line 186: The `get` function returns an optional, so you can use the or functions instead of repeating the code

mefellows: aha!

rust/pact_verifier/src/messages.rs
```diff
@@ -42,11 +43,12 @@ pub async fn verify_message_from_provider<F: RequestFilterExecutor>(
   };
   match make_provider_request(provider, &message_request, options, client).await {
     Ok(ref actual_response) => {
+      log::debug!("actual response = {:?}", actual_response);
```
TimothyJones, line 46: I think `response` might be a bit generic - should we use "internal payload" or something? Otherwise in some cases the logs are going to say "actual response: { ....}" and then "actual message { .... }" which will only be different if you have metadata.

(aside - maybe we should log metadata to `debug` and this whole payload to `trace`, since the message is already logged below?)

rust/pact_verifier/src/messages.rs
```diff
@@ -119,3 +121,29 @@ fn display_result(body_result: ANSIGenericString<str>, metadata_result: Vec<(Str
   }
   println!("      has a matching body ({})", body_result);
 }
+
+fn extract_metadata(actual_response: &Response) -> HashMap<String, String>{
+  let mut metadata = hashmap!{
+    "contentType".into() => actual_response.lookup_content_type().unwrap_or_default(),
+    "Content-Type".into() => actual_response.lookup_content_type().unwrap_or_default()
```
uglyog, line 128: Just use `contentType` here

rust/pact_verifier/src/messages.rs
```diff
@@ -119,3 +121,29 @@ fn display_result(body_result: ANSIGenericString<str>, metadata_result: Vec<(Str
...
+  };
+
+  match actual_response.body.clone() {
+    OptionalBody::Present(ref body, _) => {
+      let m: MessageMetadata = serde_json::from_slice(body).unwrap_or(MessageMetadata{ metadata: hashmap!{} });
+      log::debug!("message metadata found in response, appending = {:?}", m);
```
TimothyJones, line 134: I think this is a `trace` level log.

mefellows:
> I think contents and metadata are too generic, so if we switch behaviour on those names then we'll do the wrong thing sometimes.

They represent the external facing model of a message pact, but I'm open to other suggestions.

uglyog:
I think the idea was to not impact existing users. Now it is just looking for an additional field, but a future well formed payload is better

rust/pact_verifier/src/messages.rs
mefellows: OK. I meant to raise that in the request as well. Can we assume that any consumer adding `Content-Type` as an expectation is satisfied by this?

rust/pact_verifier/src/messages.rs
```diff
@@ -42,11 +43,12 @@ pub async fn verify_message_from_provider<F: RequestFilterExecutor>(
   };
   match make_provider_request(provider, &message_request, options, client).await {
     Ok(ref actual_response) => {
+      log::debug!("actual response = {:?}", actual_response);
+
+      let metadata = extract_metadata(actual_response);
       let actual = Message {
         contents: actual_response.body.clone(),
```
TimothyJones, line 50: See comment directly on the PR - this doesn't match the draft implementation on the client (which does match your proposed improvement ;) ).

rust/pact_verifier/src/messages.rs
uglyog: I think this is a `debug` level log.

TimothyJones:
> They represent the external facing model of a message pact, but I'm open to other suggestions.

I think I'm missing something here. What is the reason that `__pactMessageMetadata` doesn't work?

mefellows:
> I think I'm missing something here. What is the reason that `__pactMessageMetadata` doesn't work?

Because if you return a stream of bytes, or a single JSON primitive as the actual message, you can't add another field to it.

So you need a container to wrap the actual response from the message producer.

Your proposal works, but then I'm not sure why we need the `__` prefix - it's already a Pact specific payload then.

rust/pact_verifier/src/messages.rs
mefellows: yep. I think I'm going to make it accept an object that wraps both the message and the metadata, or fallback to current behaviour. That way we don't break existing clients.

rust/pact_verifier/src/messages.rs
mefellows: I'm going to move this into the `make_provider_request` method alongside the other debugging statements. That way it will be logged for all uses of this important method.

rust/pact_verifier/src/messages.rs
TimothyJones: Following the guidelines we were talking about last week [recorded here](https://github.com/pact-foundation/pact-specification#logging-in-a-pact-library), I think this log is more for maintainers than it is for users (because the whole metadata structure is already logged elsewhere). 

If I'm reading it correctly - which I might not be - this is logging the construction of the metadata structure/object/hash (whatever it's called in rust), which is probably only useful when debugging pact, not when a user is debugging their tests.

Another test is the language in the log message -a user doesn't know what "appending" means in this context, but the maintainer(s) of this code do.

I think the best experience for users would have the metadata logged once, using a log message that requires no extra context.

TimothyJones:
> Because if you return a stream of bytes, or a single JSON primitive as the actual message, you can't add another field to it.
> 
> So you need a container to wrap the actual response from the message producer.

Yes, that's the proposed implementation - I don't think we should be adding extra fields to a user's payload.

> 
> Your proposal works, but then I'm not sure why we need the `__` prefix - it's already a Pact specific payload then.

I suppose `pactMessageMetadata` works just as well, the `__` is only to increase the chance that we're not going to collide with other fields. I think `metadata` doesn't work, because we don't know if we're getting the pact-specific payload or just the raw message content.

rust/pact_verifier/src/messages.rs
mefellows: Let's leave this for another pass through. I did update another to trace though.

I think related to this, we should read through the info and debug logs for key scenarios and ask ourselves if it presents a coherent and clear view of what's happening. This way, we aren't just changing log levels, but are proactively giving the best experience for users to self diagnose, and we get good feedback if a bug request is raised.

rust/pact_matching/src/models/message.rs
```diff
@@ -301,7 +298,7 @@ impl HttpPart for Message {
     self.metadata.iter().find(|(k, _)| {
       let key = k.to_ascii_lowercase();
       key == "contenttype" || key == "content-type"
-    }).map(|(_, v)| v.clone())
+    }).map(|(_, v)| v[0].as_str().unwrap_or_default().to_string())
```
uglyog, line 301: This looks strange to me. Is the intent to convert the Value to a string? There is a util function `json_to_string` which does that

rust/pact_matching/src/models/message.rs
```diff
@@ -269,7 +266,7 @@ impl Message {
         let key = k.to_ascii_lowercase();
         key == "contenttype" || key == "content-type"
       }) {
-        Some((_, v)) => ContentType::parse(v.as_str()).ok(),
+        Some((_, v)) => ContentType::parse(v.as_str().unwrap_or_default()).ok(),
```
uglyog, line 269: Can use `json_to_string` function

uglyog (approved):

rust/pact_matching/src/models/message.rs
mefellows: Thanks. I'll check to see other uses of that now too.

rust/pact_matching/src/lib.rs
```diff
@@ -1420,7 +1423,7 @@ pub fn match_message_metadata(
         },
         None => {
           result.insert(key.clone(), vec![Mismatch::MetadataMismatch { key: key.clone(),
-            expected: value.as_str().unwrap_or_default().to_string(),
+            expected: json_to_string(&value),
```
mefellows, line 1426: @uglyog this look OK to you? (not introduced in this PR but the tests pass and my complex metadata check with Pact Go also passes)

rust/pact_matching/src/models/message.rs
mefellows: Aha! I knew I'd  seen that pattern elsewhere :P 

Screen Shot 2021-07-06 at 2 31 48 pm

I've found 3 other places that used this same approach before and have changed. Could you please double check?

The tests pass and my complex metadata check with Pact Go also passes, but good to get a second pair of eyes on it.

rust/pact_matching/src/lib.rs
uglyog: LGTM

PR #129, opened 2021-07-30
refactor: Introduce DocPath struct for path expressions

cstepanian:
- DocPaths are eagerly parsed into tokens, instead of lazily at usage time.
- Headers and Query parameters are parsed into DocPaths, and the first Field token is extracted to recover the plain string.

- Many functions now take `DocPath` instead of `&str`.  This might be considered a breaking change.
- Errors when parsing DocPaths will now be fatal errors during the parsing of Pact JSON structures.  This might be considered a breaking change.

- Moved `fn obj_key_for_path` from the `pact_consumer` crate into `path_exp`.
- Deleted now-obsolete tests for generators handling invalid expressions.
- (Code Style) Split function signatures and invocations into multiple lines and added trailing commas, where changes were already being made.

### Notice

### DIFF
--- a/rust/pact_matching/src/lib.rs
+++ b/rust/pact_matching/src/lib.rs
@@ -356,7 +356,6 @@ use pact_models::http_parts::HttpPart;
 use pact_models::interaction::Interaction;
 use pact_models::json_utils::json_to_string;
 use pact_models::matchingrules::{Category, MatchingRule, MatchingRuleCategory, RuleList};
-use pact_models::path_exp::{calc_path_weight, path_length};
 use pact_models::PactSpecification;
 use pact_models::request::Request;
 use pact_models::response::Response;
@@ -436,16 +435,16 @@ impl MatchingContext {
   /// If there is a wildcard matcher defined at the path in this context
   #[deprecated(since = "0.8.12", note = "Replaced with values matcher")]
   pub fn wildcard_matcher_is_defined(&self, path: &[&str]) -> bool {
-    !self.matchers_for_exact_path(path).filter(|&(val, _)| val.ends_with(".*")).is_empty()
+    !self.matchers_for_exact_path(path).filter(|&(val, _)| val.is_wildcard()).is_empty()
   }
 
   fn matchers_for_exact_path(&self, path: &[&str]) -> MatchingRuleCategory {
     match self.matchers.name {
       Category::HEADER | Category::QUERY => self.matchers.filter(|&(val, _)| {
-        path.len() == 1 && path[0] == *val
+        path.len() == 1 && Some(path[0]) == val.first_field()
       }),
       Category::BODY => self.matchers.filter(|&(val, _)| {
-        calc_path_weight(val, path).0 > 0 && path_length(val) == path.len()
+        val.matches_path_exactly(path)
       }),
       _ => self.matchers.filter(|_| false)
     }
@@ -1520,10 +1519,12 @@ pub fn generate_request(request: &Request, mode: &GeneratorTestMode, context: &H
   if !generators.is_empty() {
     debug!("Applying header generators...");
     apply_generators(mode, &generators, &mut |key, generator| {
-      if let Some(ref mut headers) = request.headers {
-        if headers.contains_key(key) {
-          if let Ok(v) = generator.generate_value(&headers.get(key).unwrap().clone(), context, &DefaultVariantMatcher.boxed()) {
-            headers.insert(key.clone(), v);
+      if let Some(header) = key.first_field() {
+        if let Some(ref mut headers) = request.headers {
+          if headers.contains_key(header) {
+            if let Ok(v) = generator.generate_value(&headers.get(header).unwrap().clone(), context, &DefaultVariantMatcher.boxed()) {
+              headers.insert(header.to_string(), v);
+            }
           }
         }
       }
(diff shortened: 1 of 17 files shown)


cstepanian:
It seems Clippy is raising an issue where I've derived `PartialEq` but manually implemented `Hash`.
I'll add a manual implementation of `PartialEq` to fix this.

rust/pact_matching/src/lib.rs
```diff
@@ -436,16 +435,16 @@ impl MatchingContext {
...
   pub fn wildcard_matcher_is_defined(&self, path: &[&str]) -> bool {
-    !self.matchers_for_exact_path(path).filter(|&(val, _)| val.ends_with(".*")).is_empty()
+    !self.matchers_for_exact_path(path).filter(|&(val, _)| val.is_wildcard()).is_empty()
   }
 
   fn matchers_for_exact_path(&self, path: &[&str]) -> MatchingRuleCategory {
     match self.matchers.name {
       Category::HEADER | Category::QUERY => self.matchers.filter(|&(val, _)| {
-        path.len() == 1 && path[0] == *val
+        path == ["$"] && val.is_root()
```
uglyog, line 444: This should be something like `path.last() == val`

rust/pact_models/src/matchingrules.rs
```diff
@@ -1065,10 +1071,10 @@ impl MatchingRules {
     let category = category.into();
     match category {
       Category::BODY => self.rules_for_category(Category::BODY).map(|category| category.filter(|&(val, _)| {
-        calc_path_weight(val, path).0 > 0 && path_length(val) == path.len()
+        val.matches_path_exactly(path)
       })),
       Category::HEADER | Category::QUERY => self.rules_for_category(category.clone()).map(|category| category.filter(|&(val, _)| {
-        path.len() == 1 && path[0] == *val
+        path == &["$"] && val.is_root()
```
uglyog, line 1077: This is the same as before, we want to check that the path contains the given header or query parameter name

rust/pact_models/src/path_exp.rs
```diff
@@ -35,39 +43,204 @@ fn matches_token(path_fragment: &str, path_token: &PathToken) -> usize {
...
+pub struct DocPath {
+  path_tokens: Vec<PathToken>,
+  expr: String,
+}
+
+impl DocPath {
+  pub fn new(expr: impl Into<String>) -> anyhow::Result<Self> {
+    let expr = expr.into();
+    let path_tokens = parse_path_exp(&expr)
+      .map_err(|e| anyhow!(e))?;
```
uglyog, line 58: If we update `parse_path_exp` to return an anyhow Result, we won't need to map anything here

uglyog: I see you have created a function for this: `first_field`

rust/pact_models/src/matchingrules.rs
uglyog: I see you have created a function for this: `first_field`

uglyog (changes requested):
Looking good. There is only 2 places were we should use the `first_field` function.

rust/pact_matching/src/lib.rs
uglyog: I see you have created a function for this: `first_field`

cstepanian: I'd be happy to make the change, but I don't (yet!) understand the semantics of this check (whether in the original code, or the change you are suggesting now).

When I read this code I had assumed that `path` here would always start with the Root token `$`, so the check that `path.len() == 1` was an indirect way to say that `path == ["$"]`. But now I'm not so sure that `path` always start with the Root token, since I found code in `pact_matching/src/lib.rs` where, depending on the type of thing being matched, the path could be empty (`match_path`, `match_status`) or start directly with a key (`match_metadata_value`). Notably, both header names and query parameters seem to get preceded by the Root token though.
All that to say, I thought the original code was checking that both `path` and `val` were just the Root token, which is why I rewrote it this way.

Is it possible for `path` to actually just be a header name or query parameter here?
In that case should this line actually be `path.len() == 1 && Some(path[0]) == val.first_field()`?

rust/pact_models/src/path_exp.rs
cstepanian: Ok, happy to do that. I had initially avoided touching `parse_path_exp` to keep this already huge diff from endlessly expanding :)

rust/pact_matching/src/lib.rs
uglyog: No, the root token is only used with bodies. For query parameters and headers, the path will only have the single value with the parameter name. So, checking the length would be correct. If there is more than one value in the path, it will never match anything.

rust/pact_models/src/path_exp.rs
uglyog: That's a fair comment. We can change that later.

rust/pact_matching/src/lib.rs
cstepanian: Ok, that's interesting. I trust that you know what you're talking about and I pushed a new version with this change.

But I must admit that I could not figure out where in the codebase the `path` Vec was being assembled that would lead to this function being called where `path == ["HEADER"]`, for example. I see tests for `values_matcher_defined`, but those only involve Body matchers and paths that start with `$`. I see that `values_matcher_defined` is called by `compare_maps_with_matchingrule`, which is called by `compare_maps` in `pact_matching/src/json.rs`. All the `compare_*` functions in that file seem to end up being called by `match_json`, which starts with a path of `$`.

So from what I can tell right now, it seems that the matching logic for Headers and Queries don't ever call this function, and there is no path through the codebase that would end up triggering the case `Category::HEADER | Category::QUERY` on this `match` statement. Is this part of the public API? In theory, clients can call some `pub` functions (namely `wildcard_matcher_is_defined` and `values_matcher_defined`) with their own `path` value, that makes this not strictly dead code. But I don't understand why a client would call those functions directly.

uglyog: You're right, this function is only used with collections, which will never be used with headers or query parameters. So we could actually remove that code.

PR #130, opened 2021-08-09
feat: support new selectors

mefellows:


mefellows:
Should we just remove the typing Ron?

uglyog:
What do you mean by `current typing`? As in the model struct?

mefellows:
Yeah, just wondering if we're better off just allowing any `serde_json::Value`, to save us having to constantly update the model given the validation is all in the broker anyway.

uglyog:
That feels like a bad thing to do, we should at least validate that it is correctly formed JSON so we can error early if it is invalid. I don't want to pass through garbage.

mefellows:
Sure, I just meant we don't validate the fields. Essentially, any valid JSON would be acceptable. The broker will come back with messages to tell us what is bad about the JSON.

What do you think though - keeping the types feels nicer to me, especially for anybody wanting to integrate into Rust because the types are clearer. It's not a lot of work adding new selectors.

Another thought I had was to make it an Enum which accepts the typed object, or fallback to key/value pairs (so that essentially you can just override).

That gives you the flexibility to put any valid JSON in, but still use real types if you know what you're doing. But might be overly messy.

uglyog:
I prefer keeping the types. 

Not sure about the enum, I haven't looked at the serde support, but I assume you are thinking about adding an unknown enum  with a key/value. That might work.

mefellows:
Let's keep the types then. PR should be good to go!

uglyog (approved):

PR #131, opened 2021-08-12
chore: release m1 arm package for new Mac hardware

mefellows:
See previous build, I created a step in the build pipeline to generate an M1 lib and upload the artifact: https://github.com/pact-foundation/pact-reference/actions/runs/1120206726

It does this by cross compiling, which can be done on Mac OS 10.5 and 11+.

Tested this with @TimothyJones on his M1 laptop last night and it seemed to work.

I think this is all that is required to publish?


TimothyJones:
High fives all around!!!

mefellows:
I don't know why the build is failing, because I haven't touched the build stage 🤷

.github/workflows/release.yml
```diff
@@ -18,6 +18,7 @@ jobs:
           profile: minimal
           toolchain: stable
           override: true
+          target: aarch64-apple-darwin
```
uglyog, line 21: Does this set the target or add it to the default target?

uglyog: Ok, it `Additionally install specified target for this toolchain`

uglyog (approved):

uglyog:
I have a quicktest test to test the integer validation, and every once in a while it finds a character that is convertible to a number but is not recognised by the standard lib as something that can be converted to a number :-D
```
thread 'verifier::args::test::validates_integer_value' panicked at '[quickcheck] TEST FAILED. Arguments: ("⁰")'
```

uglyog:
Safe to ignore it

PR #135, opened 2021-08-22
feat(ffi verifier): added crude method to extract CLI args, and make available via FFI

mikegeeves:


mikegeeves (commented):
Ahh in my haste I created here instead of just in my fork. Cancel! Cancel!

PR #138, opened 2021-08-31
chore: upgrade tree_magic_mini to 3.0.1

audunhalland:
This upgrades all of `nom` to `7.0.0`.

In order to pass tests, the `with-gpl-data` feature needed to be added. `tree_magic_mini` is GPL-licensed, because [the MIME database is](https://github.com/mbrubeck/tree_magic#licensing-and-the-mime-database). Which in turn makes pact-reference itself GPL.

Note that tree_magic_mini is already GPL-licensed at [version 2](https://github.com/mbrubeck/tree_magic/blob/mini/CHANGELOG.md).

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -36,12 +36,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "28ae2b3dec75a406790005a200b1bd89785afc02517a00ca99ecfe093ee9e6cf"
 
 [[package]]
-name = "arrayvec"
-version = "0.5.2"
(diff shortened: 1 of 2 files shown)


uglyog:
I'm just going to get feedback from the other maintainers if forcing to GPL license is a problem. It might be better to not include the MIME database, which means Windows uses will not get the content type detection for binary data. I don't think lots of people use that anyway.

audunhalland:
Mime detection is not important to me, getting rid of nom 6 is. This PR changes nothing about the licensing issue.. MIME removal could be handled in another PR?

uglyog:
I'll sort it out when I get a chance to merge this

PR #139, opened 2021-09-08
feat: add verifier ffi function set verification options

tienvx:


rust/pact_ffi/src/verifier/mod.rs
```diff
@@ -134,7 +134,7 @@ ffi_fn! {
     /// All string fields must contain valid UTF-8. Invalid UTF-8
     /// will be replaced with U+FFFD REPLACEMENT CHARACTER.
     ///
-    fn pactffi_verifier_update_provider_state(
+    fn pactffi_verifier_set_provider_state(
```
tienvx, line 137: It's my mistake. It should follow naming pattern from `pactffi_verifier_set_provider_info`

rust/pact_ffi/src/verifier/mod.rs
```diff
@@ -147,6 +147,46 @@ ffi_fn! {
...
+      provider_version: *const c_char,
+      build_url: *const c_char,
+      disable_ssl_verification: c_uchar,
+      request_timeout: c_ulong,
+      provider_tags: *const *const c_char,
+      provider_tags_len: c_ushort
+    ) {
+      let handle = as_mut!(handle);
+      let provider_version = safe_str!(provider_version);
+      let build_url = safe_str!(build_url);
```
tienvx, line 173: I tested with `provider_version = null` or `build_url = null`, it stop right here, not running further.

Is it okay? If not, I think we must use `if_null`. Can you suggest default values for `provider_version` and `build_url` if so?

Same for `url` in `pactffi_verifier_set_provider_state`

rust/pact_ffi/src/verifier/mod.rs
```diff
@@ -147,6 +147,46 @@ ffi_fn! {
...
+      let tags = if !provider_tags.is_null() && provider_tags_len > 0 {
+        let mut tags = Vec::with_capacity(provider_tags_len as usize);
+        for index in 0..(provider_tags_len - 1) {
+          let tag_ptr: * const c_char = unsafe { *(provider_tags.offset(index as isize)) };
+          tags.push(safe_str!(tag_ptr).to_string());
+        }
+        tags
+      } else {
+        vec![]
+      };
```
tienvx, line 184: I can not extract this code into new function to avoid duplication. I am keep getting this error on build:

```
    |
569 |       tags.push(safe_str!(tag_ptr).to_string());
    |                 ^^^^^^^^^^^^^^^^^^ expected struct `Vec`, found enum `Result`
    |
```

rust/pact_ffi/src/verifier/mod.rs
uglyog: `provider_version` is required, so if that is not set, we should error at that point.

`build_url` is optional, so we should check for null and set a default then. 

For the provider state url, I think that can be optional. The provider states will just be skipped then.

uglyog: Just checked the `safe_str` macro, and it will panic if the pointer is null, which is probably not the behaviour we want.

rust/pact_ffi/src/verifier/mod.rs
uglyog: I'll need to see the code you extracted to understand this error

tienvx: This is the extracted function:
```rust
fn get_tags(provider_tags: *const *const c_char, provider_tags_len: c_ushort) -> Vec<String>
{
  let tags = if !provider_tags.is_null() && provider_tags_len > 0 {
    let mut tags = Vec::with_capacity(provider_tags_len as usize);
    for index in 0..(provider_tags_len - 1) {
      let tag_ptr: * const c_char = unsafe { *(provider_tags.offset(index as isize)) };
      tags.push(safe_str!(tag_ptr).to_string());
    }
    tags
  } else {
    vec![]
  };
  tags
}
```

rust/pact_ffi/src/verifier/mod.rs
uglyog: Use the `if_null` function for `build_url`, and set the default to the empty string.

For `provider_version`, it is mandatory, so you need to add an error block as a second block that returns an error code when the `safe_str!` panics. See `pactffi_message_set_description` for an example on how to do that.

rust/pact_ffi/src/verifier/mod.rs
uglyog: I don't know why that is happening, I get the same issue. The `safe_str` macro is doing something odd in that context.

Changing the function to the following works:
```rust
fn get_tags(provider_tags: *const *const c_char, provider_tags_len: c_ushort) -> Vec<String> {
  if !provider_tags.is_null() && provider_tags_len > 0 {
    let mut tags = Vec::with_capacity(provider_tags_len as usize);
    for index in 0..(provider_tags_len - 1) {
      let tag_ptr: *const c_char = unsafe { *(provider_tags.offset(index as isize)) };
      let tag = if_null(tag_ptr, "");
      if !tag.is_empty() {
        tags.push(tag.to_string());
      }
    }
    tags
  } else {
    vec![]
  }
}
```

rust/pact_ffi/src/verifier/mod.rs
tienvx: Thanks. I learnt something new. Updated.

rust/pact_ffi/src/verifier/mod.rs
tienvx: It works. Thanks

PR #149, opened 2021-09-20
feat: support native TLS. Fixes #144

### LINKED ISSUE #144: feature: support disabling and providing custom SSL certificates
We currently support it for the provider but not the broker itself. I think we should:

1. Support disabling SSL for the broker interactions
2. Support specifying custom certificate paths for the broker and the verifier (ideally supporting the OpenSSL convention of the `SSL_CERT_FILE` and `SSL_CERT_DIR` environment variables).

See https://github.com/pact-foundation/pact-js/issues/736 for background, it may already be partially supported through the reqwest dependencies.

mefellows:
Added some docs to explain the benefits to the verifier.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1611,6 +1611,7 @@ dependencies = [
  "percent-encoding",
  "pin-project-lite",
  "rustls",
+ "rustls-native-certs",
  "serde",
  "serde_json",
  "serde_urlencoded",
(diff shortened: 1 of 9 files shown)


mefellows:
Tested on OSX - see also https://github.com/pact-foundation/pact-reference/issues/144.

Much better than having to manually set environment variables!

Let me know if OK to merge?

Issue #151, opened 2021-09-29
Add support for publishing verification results with branches


bethesque:
This change allows the provider version associated with verification results to be created with a branch.

* Add a "branch" option to the API exposed to the users
* When fetching the pacts for verification, set the branch as `providerVersionBranch` in the request.
* When publishing the verification results, before creating the tags, send a `PUT` using the `pb:branch-version` relation from the provider resource (with params `branch` and `version`), or the `pb:pacticipant-branch-version` relation from the index (with params `pacticipant`, `branch` and `version`).
* If the branch is set, and the broker does not support branches (ie. the relations are not there), raise an appropriate error (or print warning - which ever works best for rust) indicating that the broker does not support branches, and they should update to Pact Broker version 2.86.0 or later for branch support.

Sorry to add yet another call to the results publishing! I imagine the rust code for publishing verification results looks about as bad as the ruby code does now 😆  There will be an "all in one" verification results publishing endpoint coming at some stage, but I don't want to wait for that to be done before we can get people using branches.

mefellows:
We already support `branch` in the selectors, does this mean it will automatically flow in the verification results publish, or does that need to be passed in to that API somehow?

There is a tiny Pact CLI in rust that supports publishing pacts, albeit I'm not sure we want it personally. We have enough things to maintain!

bethesque:
> We already support branch in the selectors, does this mean it will automatically flow in the verification results publish, or does that need to be passed in to that API somehow?

That's for the consumer version. The change is to set the branch for the provider version. But that reminds me, it also needs to be sent in the 'pacts for verification' call. I'll update the original comment.

mefellows:
I've started picking this up. Note to myself on testing:

> publish a pact with a branch, use consumer version selectors with a branch, publish a verification with a branch, then look at the matrix page

mefellows:
@bethesque so I have a consumer where I have published with a branch:

```
pact-broker publish ${PWD}/pacts --consumer-app-version ${GIT_COMMIT} --branch ${GIT_BRANCH}
...
Created version aa2dccc+1632918640 of pactflow-example-consumer (branch master)
Publishing pactflow-example-consumer/pactflow-example-provider-dredd pact to pact broker at https://testdemo.pactflow.io
The latest version of this pact can be accessed at the following URL:
https://testdemo.pactflow.io/pacts/provider/pactflow-example-provider-dredd/consumer/pactflow-example-consumer/latest
```

[...]

PR #152, opened 2021-09-29
feat: support provider branches

mefellows:
Just an FYI that it's in progress.

See also #151 - seems to work as expected.


rust/pact_ffi/src/verifier/mod.rs
```diff
@@ -369,11 +369,18 @@ ffi_fn! {
       enable_pending: c_uchar,
       include_wip_pacts_since: *const c_char,
       provider_tags: *const *const c_char,
-      provider_tags_len: c_ushort
+      provider_tags_len: c_ushort,
+      provider_branch: *const c_char
```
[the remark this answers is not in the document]
uglyog: It's probably ok because not many people are using that method yet, but it is getting too big (too many parameters).

rust/pact_verifier/src/lib.rs
```diff
@@ -90,11 +92,11 @@ impl Display for PactSource {
       PactSource::BrokerUrl(ref provider_name, ref broker_url, _, _) => {
           write!(f, "PactBroker({}, provider_name='{}')", broker_url, provider_name)
       }
-      PactSource::BrokerWithDynamicConfiguration { ref provider_name, ref broker_url,ref enable_pending, ref include_wip_pacts_since, ref provider_tags, ref selectors, ref auth, links: _ } => {
+      PactSource::BrokerWithDynamicConfiguration { ref provider_name, ref broker_url,ref enable_pending, ref include_wip_pacts_since, ref provider_branch, ref provider_tags, ref selectors, ref auth, links: _ } => {
```
uglyog, line 95: LOL old code from long ago when I didn't know what I was doing. All the `ref` keywords are no longer necessary.

uglyog (approved):

rust/pact_verifier/src/lib.rs
mefellows: noted

mefellows:
Happy to merge then or want me to make those changes?

uglyog:
> Happy to merge then or want me to make those changes?

Which changes? If it's the note about the ref keyword, don't worry about that, it is just unnecessary in the source code (Rust now automatically makes the fields references if the outer thing is a reference)

PR #167, opened 2021-12-06
fix(verifier): the state_change_teardown option wasn't used

mikegeeves:


rust/pact_verifier/src/callback_executors.rs
```diff
@@ -67,6 +67,9 @@ pub trait ProviderStateExecutor {
     setup: bool,
     client: Option<&reqwest::Client>
   ) -> anyhow::Result<HashMap<String, Value>>;
+
+  /// If a teardown call for the Executor should be performed
+  async fn teardown(self: Arc<Self>)-> bool;
```
uglyog, line 72: This doesn't need to be async, it is only returning a boolean value.

uglyog (changes requested):

PR #173, opened 2022-01-25
feat(ffi)!: Separate verification and publishing options

adamrodger:
This change makes the verifier FFI easier to work with because the publishing options are split from the verification options.

The verification options should apply to all pact sources (such as file and URL) because they contain global settings such as request timeouts and callbacks.

However, the publish options only make sense when a Pact Broker is configured. This made the FFI tricky to use from library code (in particular from PactNet) because that meant you could only set the global verification options when using a Pact Broker source, instead of when using any source. This change will allow that happen properly in future.

As an aside, I didn't find any unit tests to update at all for this, which is a bit concerning.

### DIFF
--- a/rust/pact_ffi/src/verifier/handle.rs
+++ b/rust/pact_ffi/src/verifier/handle.rs
@@ -5,7 +5,7 @@ use std::sync::Arc;
 use log::debug;
 
 use pact_models::prelude::HttpAuth;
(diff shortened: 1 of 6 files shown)


mefellows:
Awesome, thanks Adam!

mefellows:
RE: tests, for previous FFI changes I've made, I've created an acceptance suite. I haven't looked at the FFI code here in detail, but for the functions exposed as the c interface they lend themselves more to this style of test. There is likely other glue code below this that could be tested differently.

uglyog:
The only type of tests I would expect is testing behavior when a NULL pointer is passed in. They should just be wrapping other functionality that should have the unit tests.

PR #175, opened 2022-01-26
fix: provider request timeout should be > 16bit integers. 

mefellows:
Fixes https://github.com/pact-foundation/pact-js/issues/761

Alternatively, could just make `integer_value` a 64bit number and not worry too much about restricting the port range.

### DIFF
--- a/rust/pact_ffi/src/verifier/args.rs
+++ b/rust/pact_ffi/src/verifier/args.rs
@@ -1,8 +1,12 @@
 use clap::{App, Arg};
 use regex::Regex;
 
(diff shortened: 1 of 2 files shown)


adamrodger:
The new FFI verifier handles approach also has it as a 16bit number, so we may want to update there also to match.

mefellows:
ah, thanks Adam!

mefellows:
It looks like the port is a `u16` but the request timeout is a `u_64`, courtesy of #173. Unless I'm missing something?

PR #177, opened 2022-01-28
Feat/capture verifier output

uglyog:
I updated the verifier code to capture all the output and also the results. There are two new FFI functions:

`pactffi_verifier_output` - This will return the output that is written to standard out
`pactffi_verifier_json` - This will return the verification result as JSON

I have also added an argument to the verifier CLI to write the results to a JSON file.

image

That generates the following file:

```json
{
  "errors": [
    {
      "interaction": "Verifying a pact between test_consumer and test_provider Given test state - test interaction",
      "mismatch": {
        "interactionId": "",
        "message": "error sending request for url (http://localhost/): error trying to connect: tcp connect error: Connection refused (os error 111)",
        "type": "error"
      }
    }
  ],
  "notices": [],
  "output": [
    "\nVerifying a pact between \u001b[1mtest_consumer\u001b[0m and \u001b[1mtest_provider\u001b[0m",
    "",
    "  test interaction",
    "      \u001b[31mRequest Failed - error sending request for url (http://localhost/): error trying to connect: tcp connect error: Connection refused (os error 111)\u001b[0m",
    "",
    "\nFailures:\n",
    "1) Verifying a pact between test_consumer and test_provider Given test state - test interaction - error sending request for url (http://localhost/): error trying to connect: tcp connect error: Connection refused (os error 111)\n",
    "\nThere were 1 pact failures\n"
  ],
  "pendingErrors": [],
  "result": false
}
```

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1498,7 +1498,7 @@ dependencies = [
 
 [[package]]
 name = "pact_verifier"
-version = "0.12.5"
+version = "0.13.0"
 dependencies = [
  "ansi_term 0.12.1",
  "anyhow",
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -52,9 +52,9 @@ dependencies = [
 
 [[package]]
 name = "async-compression"
-version = "0.3.8"
+version = "0.3.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5443ccbb270374a2b1055fc72da40e1f237809cd6bb0e97e66d264cd138473a6"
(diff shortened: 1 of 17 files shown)


mefellows:
Awesome! I'll take a look over the weekend.

Adam made a good point about the ANSI output - does the library auto-detect if it can print colours etc.?

uglyog:
> Adam made a good point about the ANSI output - does the library auto-detect if it can print colours etc.?
Looks like for Windows it has to be enabled, because it probably uses the TERM environment variables. See https://github.com/ogham/rust-ansi-term#basic-usage

PR #183, opened 2022-02-28
Feat/custom headers

### LINKED ISSUE #182: Feature: Implement CustomHeaders handling
When writing provider tests in pact-net 3.x.x  (that used the Ruby Implementation) we had the option to define CustomHeader manipulation. That was a useful feature, because in that way we could easily handle authentication token being invalid without having to use middleware to change requests on the fly.

Such implementation would then be called in other implementations that uses this one as the base for.

Thanks in advance

uglyog:
This adds a new function to set custom headers: 
```
pactffi_verifier_add_custom_header(
      handle: *mut handle::VerifierHandle,
      header_name: *const c_char,
      header_value: *const c_char
    )
```

Resolves #182

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1535,6 +1535,7 @@ dependencies = [
  "clap",
  "expectest",
  "log",
+ "maplit",
  "pact_models 0.2.7",
  "pact_verifier",
  "regex",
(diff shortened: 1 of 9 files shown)


mefellows:
Awesome, Thanks Ron. Signature looks good, I'll take a look at implementation tomorrow

mefellows (approved):

adamrodger (approved):

Issue #181, opened 2022-02-18
Can’t publish verification results with pact_verifier and PactSource::BrokerWithDynamicConfiguration


kageru:
I’m currently using pact_verifier 0.13.1 to verify pacts from a pact broker. The pacts are received and verified fine, but it seems like the result publishing step is simply skipped.

After looking at the code, it seems to me that even with `PublishOptions` present, results are only ever published when the pact source is an instance of `PactSource::BrokerUrl` as implemented in https://github.com/pact-foundation/pact-reference/blob/pact_verifier-v0.13.1/rust/pact_verifier/src/lib.rs#L1032, which leads me to my first question:

Should publishing be possible when using a `PactSource::BrokerWithDynamicConfiguration` (since that contains all the fields of `BrokerUrl` and more), and if yes, would you be interested in a PR that implements that?

**Edit**: the second question has been resolved. No need to keep reading.

Secondly, even when using a normal `BrokerUrl` as my pact source (see code example below), the broker doesn’t show pacts as verified. I’ve seen some logging inside of the library that could help me debug this, but I don’t know how to redirect the `info!` and `error!` logs to stdout. Simply using `--nocapture` in my tests does nothing.

```rs
use pact_models::v4::http_parts::HttpRequest;
use pact_verifier::{
    callback_executors::{HttpRequestProviderStateExecutor, RequestFilterExecutor},
    FilterInfo, PactSource, ProviderInfo, PublishOptions, VerificationOptions,
};
use std::sync::Arc;

#[tokio::test]
async fn test() {
    let result = pact_verifier::verify_provider_async(
        ProviderInfo {
            name: "our-provider".into(),
            ..Default::default()
        },
        vec![PactSource::BrokerUrl("our-provider".into(), "https://my-pact-broker.com".into(), None, vec![])],
        FilterInfo::None,
        vec!["their-consumer".into()],
        &VerificationOptions {
            request_filter: Some(Arc::new(Filter)),
            ..Default::default()
        },
        Some(&PublishOptions {
            provider_version: Some("0.1.0".into()),
            provider_tags: vec!["master".into()],
            provider_branch: Some("master".into()),
            build_url: None,
        }),
        &Arc::new(HttpRequestProviderStateExecutor::default()),
        None,
    )
    .await
    .unwrap();
    // This assertion passes.
    // When printing the entire `result` object, I can see that it verified
    // all pacts I can see on the broker, so that part definitely works.
[...]

PR #194, opened 2022-05-17
Allow BrokerWithDynamicConfiguration to publish results

kageru:
As explained in #181, this should be possible.

This is my first time interacting with this codebase, so please let me know if there’s anything wrong with what I’ve implemented here and/or if there are other ways to check for regressions than `cargo test`.


rholshausen:
All good, thanks for the PR!

PR #211, opened 2022-08-03
Feat/verifier multiple transports

rholshausen:
Updates the CLI verifier to support verifying Pacts were there are different interactions (i.e HTTP app and then a different proxy for the messages).

image

### DIFF
--- a/rust/pact_ffi/src/verifier/handle.rs
+++ b/rust/pact_ffi/src/verifier/handle.rs
@@ -82,9 +82,10 @@ impl VerifierHandle {
       port: port.clone(),
       path: path.clone(),
       transports: vec![ ProviderTransport {
-        transport: scheme,
+        transport: scheme.clone(),
         port,
-        path: if path.is_empty() { None } else { Some(path) }
+        path: if path.is_empty() { None } else { Some(path) },
+        scheme: None
       } ]
     }
   }
--- a/rust/pact_ffi/src/verifier/handle.rs
+++ b/rust/pact_ffi/src/verifier/handle.rs
@@ -7,16 +7,7 @@ use pact_models::prelude::HttpAuth;
 use serde_json::Value;
 use tracing::debug;
 
-use pact_verifier::{
-  ConsumerVersionSelector,
-  FilterInfo,
-  NullRequestFilterExecutor,
-  PactSource,
-  ProviderInfo,
-  PublishOptions,
-  VerificationOptions,
-  verify_provider_async
-};
+use pact_verifier::{ConsumerVersionSelector, FilterInfo, NullRequestFilterExecutor, PactSource, ProviderInfo, ProviderTransport, PublishOptions, VerificationOptions, verify_provider_async};
 use pact_verifier::callback_executors::HttpRequestProviderStateExecutor;
 use pact_verifier::metrics::VerificationMetrics;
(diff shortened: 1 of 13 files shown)


rust/pact_verifier/src/lib.rs
```diff
@@ -1182,11 +1257,15 @@ pub async fn verify_pact_internal<'a, F: RequestFilterExecutor, S: ProviderState
       }
     };
 
+    // TODO: Update this to use V4 models
```
mefellows, line 1260: Is this something that needs addressing in this PR, or a reminder for cleanup in the future (i.e. no functional impact)?

rust/pact_verifier_cli/src/args.rs
```diff
@@ -1,28 +1,38 @@
...
   v.parse::<u16>().map(|_| ()).map_err(|e| format!("'{}' is not a valid port value: {}", v, e) )
 }
 
-fn integer_value(v: String) -> Result<(), String> {
+fn integer_value(v: &str) -> Result<(), String> {
   v.parse::<u64>().map(|_| ()).map_err(|e| format!("'{}' is not a valid integer value: {}", v, e) )
 }
 
-pub(crate) fn setup_app<'a, 'b>(program: String, version: &'b str) -> App<'a, 'b> {
+fn transport_value(v: &str) -> Result<(String, u16), String> {
```
mefellows, line 12: Is it possible for a transport to be a socket? (we don't need to tackle it now if it can be added).

mefellows (approved):
Nice one.

rust/pact_verifier_cli/src/args.rs
rholshausen: Sockets still need to use a port. But the verifier won't know how to write the request or read the response, so it will have to be done with a plugin

rust/pact_verifier/src/lib.rs
rholshausen: I decided to make it tech debt. Possibly just to annoy Beth.

bethesque: It's only tech debt if Matt made you do it.

rust/pact_ffi/src/verifier/handle.rs
```diff
@@ -83,12 +74,19 @@ impl VerifierHandle {
...
       host,
-      port: if port == 0 { None } else { Some(port) },
-      path
+      port: port.clone(),
+      path: path.clone(),
+      transports: vec![ ProviderTransport {
+        transport: scheme.clone(),
+        port,
+        path: if path.is_empty() { None } else { Some(path) },
+        scheme: None
```
Gicminos, line 88: If an `https` scheme as been passed to this function, why is it being set to `None` for this transport?

I believe this is causing pact-python to attempt testing a HTTPS provider via a `http://` URL

rholshausen: That's a good catch, I don't think it was intentional. I need to review what `None` means here. I hope it means use the default for the protocol, but maybe that logic is not working properly.

rholshausen: Confirmed the issue. If the scheme is `None`, it will use `http`.

rholshausen: Corrected in efc54d26

PR #212, opened 2022-08-12
Fix/send metrics

rholshausen:
* Fixes the metrics call by making the send metrics function async and getting the verification wait on it
* Adds a ConsumerTestRun event to the pactffi_cleanup_mock_server function.

Not sure if the pactffi_cleanup_mock_server is the best place for that event.

### DIFF
--- a/rust/pact_ffi/src/mock_server/mod.rs
+++ b/rust/pact_ffi/src/mock_server/mod.rs
@@ -53,6 +53,7 @@ use std::str::from_utf8;
 
 use anyhow::Context;
 use chrono::Local;
+use either::Either;
 use libc::c_char;
 use onig::Regex;
 use pact_models::pact::Pact;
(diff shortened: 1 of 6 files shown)


rust/pact_ffi/src/mock_server/mod.rs
```diff
@@ -419,7 +421,27 @@ pub extern fn pactffi_mock_server_mismatches(mock_server_port: i32) -> *mut c_ch
...
+      let interactions = match mock_server {
+        Either::Left(ms) => {
+          let pact = ms.pact.lock().unwrap();
+          pact.interactions().len()
+        },
+        Either::Right(ms) => ms.pact.interactions.len()
+      };
+      send_metrics(MetricEvent::ConsumerTestRun {
+        interactions,
+        test_framework: "pact_ffi".to_string(),
```
mefellows, line 434: Should these be coming from the client library, so we can disambiguate languages? I think we do this on the provider side right?

mefellows (approved):
Just had the clarification about passing through the client framework so we can track languages separately. Other than that, it looks OK to my eyes

rust/pact_ffi/src/mock_server/mod.rs
rholshausen: Yes, but there is no way to get that info from within the `pactffi_cleanup_mock_server` function.

mefellows: Right, yes that makes sense. We can always add later

Issue #217, opened 2022-09-22
pact_verifier_cli 0.9.16 panics at clap-3.2.20/src/parser/matches/arg_matches.rs:1795:13 when filtering pacts using consumer name


j3rrywan9:
1. The relevant versions of the packages you are using.
The official `pact-ref-verifier` Docker image with tag `0.9.16`: https://hub.docker.com/layers/pactfoundation/pact-ref-verifier/0.9.16/images/sha256-caf09b90140066410563eac2efd7a69f306ffb781befd2b81946aa92545ce0cc

2. The steps to recreate your issue.
a) Start a container from the official image using Docker Compose
```
version: "3.7"

services:
  pact-verifier:
    container_name: verifier
    image: "pactfoundation/pact-ref-verifier:0.9.16"
    entrypoint: "cat"
    tty: true
    environment:
      PACT_BROKER_TOKEN: "[MY_PACKFLOW_TOKEN]"
    volumes:
      - ./pacts:/verifyPacts
```
b) Run the `pact_verifier_cli` with the `--filter-consumer` option
```
docker exec --env RUST_BACKTRACE=full -it verifier /usr/local/bin/pact_verifier_cli --filter-consumer modelling-environment-data-tool --scheme http --hostname "api-services" --port 8180 --base-path / --consumer-version-selectors '{"branch": "develop"}' --broker-url https://anaplan.pactflow.io --provider-name api-services --state-change-url http://state-service:8001/state --loglevel debug
```
c) The above command ends up with an error stack like the following
```
2022-09-22T00:00:36.529668Z DEBUG main pact_verifier_cli: Pact source to verify = PactBrokerWithDynamicConfiguration(https://anaplan.pactflow.io, provider_name='api-services', enable_ending=false, include_wip_since=None, provider_tags=[], provider_branch=None, consumer_version_selectors='[ConsumerVersionSelector { consumer: None, tag: None, fallback_tag: None, latest: None, deployed_or_released: None, deployed: None, released: None, environment: None, main_branch: None, branch: Some("develop"), matching_branch: None }], auth=Token(Me3Q******************)')
thread 'main' panicked at 'Must use `Arg::allow_invalid_utf8` with `_os` lookups at `[hash: 2AFE180D94FE0B73]`', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/clap-3.2.20/src/parser/matches/arg_matches.rs:1795:13
stack backtrace:
[...]

PR #219, opened 2022-09-24
fix(pact_verifier_cli): stop using deprecated clap::parser::matches::arg_matches::ArgMatches::values_of_lossy

j3rrywan9:
Fixes: https://github.com/pact-foundation/pact-reference/issues/217


rholshausen:
Thanks for the PR!

PR #221, opened 2022-10-04
feat(ffi): support adding transports to provider

mefellows:


rust/pact_ffi/src/verifier/handle.rs
```diff
@@ -65,6 +65,25 @@ impl VerifierHandle {
...
+  ) {
+
+    let transport = ProviderTransport {
+      transport: protocol,
+      port: Some(port),
+      path: if path.is_empty() { None } else { Some(path) },
+      scheme: scheme
+    };
+
+    self.provider.transports.append(&mut vec![transport]);
```
rholshausen, line 84: There must be a function to add a new item, as this one creates a new vector, converts it into a mutable slice, and then appends that to the vector

rholshausen: Yeah, `push` is what we want to use

rholshausen:
This looks fine to me (apart from using append instead of push), so it must be an issue with the verifier function not using the transports

rust/pact_ffi/src/verifier/handle.rs
mefellows: I _swear_ this was the first thing I tried, but it seems to compile now so 🤷 . Fixed, thanks.

Issue #234, opened 2022-12-09
When using any other type than `application/json` in `pactffi_with_body` the match does not work


leonasa:
Calling the function this way:
`let json_ct = CString::new("application/vnd.schemaregistry.v1+json").unwrap();
    let json = "{\"test\":true}";
    let body = CString::new(json).unwrap();
    let result = pactffi_with_body(i_handle, InteractionPart::Request, json_ct.as_ptr(), body.as_ptr());`

we should expect to have the `interaction1.request.body.content_type() == "application/vnd.schemaregistry.v1+json"`
but it sets the default value instead. 
it is done in the function `pactffi_with_body` exactly here:
`  let body = if reqres.request.content_type().unwrap_or_default().is_json() {
            let category = reqres.request.matching_rules.add_category("body");
            OptionalBody::Present(Bytes::from(process_json(body.to_string(), category, &mut reqres.request.generators)),
                                  Some(JSON.clone()), None)
          }`

Issue #224, opened 2022-10-20
JSON formats other than application/json are not supported


mefellows:
See https://github.com/pact-foundation/pact-js/issues/963.

Given an expected POST body content type of `application/vnd.dog.farm+json`, when the actual request is made, the following error occurs:

```
  Mock server failed with the following mismatches:

        0) The following request was incorrect:

                PUT /dogs

                         1.0 Expected body with content type application/json but was application/vnd.dog.farm+json
```

It looks like all JSON-like types are e to application/json, and then compared to the incoming narrow type.

mefellows:
The latest version of Pact JS (and the core) should have this fix in it - could you please re-test?

PR #235, opened 2022-12-12
feat: allow non deafult json content-type

leonasa:


leonasa:
When using any other type than application/json in pactffi_with_body the match does not work #234

rholshausen:
Thanks for the PR!

mefellows:
Would this change help with #224?

leonasa:
@mefellows it is definitely the same thing. And it will help your case.

mefellows:
Awesome - thank you!

PR #237, opened 2022-12-17
feat: add user-agent header to the HALClient

### LINKED ISSUE #236: Requests for pact gets blocked by some firewalls due to not providing a user-agent header
When self-hosting a Pact Broker in, for example AWS or Azure, it is common to have some kind of web application firewall (WAF) in place. One use case for WAFs that seems to be fairly common, is to use the `User-Agent` HTTP header for filtering of requests ([example from AWS docs](https://aws.amazon.com/premiumsupport/knowledge-center/waf-block-http-requests-no-user-agent/)).

The `pact_verifier`/`pact_verifier_cli` on `master` does currently not set a `User-Agent` which means that requests get blocked before they reach the Pact Broker.

This also affect the Pact libraries in other languages using this library.

andrisak:
This PR adds a [`User-Agent`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent) to the `HALClient` used by the `pact_verifier` to interact with the Pact Broker.

Fixes #236.

### DIFF
--- a/rust/pact_verifier/src/pact_broker.rs
+++ b/rust/pact_verifier/src/pact_broker.rs
@@ -479,6 +479,7 @@ impl Default for HALClient {
   fn default() -> Self {
     HALClient {
       client: reqwest::ClientBuilder::new()
+        .user_agent(format!("{}/{}", env!("CARGO_PKG_NAME"), env!("CARGO_PKG_VERSION")))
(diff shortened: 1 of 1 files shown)


rholshausen:
Thanks for the PR!

PR #260, opened 2023-03-12
refactor: bump indextree, use append_value()

YoniFeng:
I added a fast path for adding new nodes with `indextree` - https://github.com/saschagrunert/indextree/pull/92.
From a glimpse at this project I'm guessing perf isn't important, but maybe you'll still be interested in this tiny refactor.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -988,9 +988,9 @@ dependencies = [
 
 [[package]]
 name = "indextree"
-version = "4.5.0"
+version = "4.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


rholshausen:
Thanks for the PR.

Performance has never been raised as an issue, I guess because it mainly deals with small file sizes. Do you happen to know what the benefit of this change would be? I assume you looked at some performance metrics to get to the point to make this change?

YoniFeng:
I don't know what the benefit would be for pact_models.
You can see the impact for JS parser building an AST [here](https://github.com/saschagrunert/indextree/pull/92#issuecomment-1455131660).

I got here after making the change in `indextree` and thought "I should look at dependents that can benefit".
`indextree` doesn't have huge dependents, and this project was at the top :)

rholshausen:
Nice! And thanks again for the PR.

PR #266, opened 2023-04-12
feat: pact_mock_server_cli osx aarch64

YOU54F:
# pact_mock_server_cli

- Add aarch64-apple-darwin target for pact_mock_server_cli
- Adds smoke test for  aarch64-apple-darwin target for pact_mock_server_cli

Bonus

- Adds smoke test for  aarch64-apple-darwin target for pact_mock_server_cli
- Adds smoke test for  aarch64-unknown-linux-gnu target for pact_verifier_cli
- Adds smoke test for  aarch64-unknown-linux-gnu target for pact_verifier_cli

##  Supported Platforms

| OS           | Architecture | Supported |
| ------- | ------------ | --------- |
| OSX        | x86_64       | ✅         |
| OSX        | aarch64 (arm)| ✅         |
| Linux    | x86_64       | ✅         |
| Linux    | aarch64 (arm)| ✅         |
| Windows    | x86_64  | ✅         | 

## Testing

CI can be run locally with [tart.run](https://tart.run/) and [cirrus-cli](https://github.com/cirruslabs/cirrus-cli) on a mac machine with `tart run` in the root folder.

Cirrus-CI account needed for cloud testing of ARM64 machines, all testable within the free tier and no credit card needed

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,31 @@
+BUILD_TEST_TASK_TEMPLATE: &BUILD_TEST_TASK_TEMPLATE
+  test_pact_mock_server_cli_script: |
+    cd rust/pact_mock_server_cli
+    cargo build --target ${BUILD_TARGET} --release
+    ../target/${BUILD_TARGET}/release/pact_mock_server_cli --help
+    rm -rf ../target/
+  test_pact_verifier_cli_script: |
+    cd rust/pact_verifier_cli
+    cargo build --target ${BUILD_TARGET} --release
+    ../target/${BUILD_TARGET}/release/pact_verifier_cli --help
+
+
+linux_arm64_task: 
+  arm_container:
+    image: rustlang/rust:nightly
+    cpu: 4
+    memory: 12G
+  env:
+    BUILD_TARGET: aarch64-unknown-linux-gnu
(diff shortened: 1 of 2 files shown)


.cirrus.yml
```diff
@@ -0,0 +1,31 @@
...
+    rm -rf ../target/
+  test_pact_verifier_cli_script: |
+    cd rust/pact_verifier_cli
+    cargo build --target ${BUILD_TARGET} --release
+    ../target/${BUILD_TARGET}/release/pact_verifier_cli --help
+
+
+linux_arm64_task: 
+  arm_container:
+    image: rustlang/rust:nightly
```
rholshausen, line 15: Is there a reason to need nightly here, I would prefer to run on stable and/or beta

YOU54F: Yeah I would have thought that too, I chose it as it was from the release flow where these get built

https://github.com/pact-foundation/pact-reference/blob/1116eff9850781c802217bc67481a84474d3dc5c/.github/workflows/release.yml#L49

Although that only appears before the FFI run (which is maybe more suspect) especially in a release run

kicked off a run against stable tag now - https://cirrus-ci.com/build/4982626734309376

PR #270, opened 2023-04-21
ci(chore) - various updates to replace deprecated action notices and red builds

YOU54F:
Hey @uglyog 

Noted the CI workflows were throwing a few deprecation notices, and the release workflow was failing on any jobs that were for the ffi, verifier cli, or mock server cli, so this PR add's a few changes to clean up the build

1. Update `actions/checkout@v2` to `actions/checkout@v3`
2. Switch deprecated `actions-rs/toolchain@v1` with `dtolnay/rust-toolchain` https://github.com/dtolnay/rust-toolchain as original unmaintained - see [issue](https://github.com/actions-rs/toolchain/issues/216) 
3. Replace `marcopolo/cargo@master` also unmaintained, and not sure using an action for this adds value based on what we are using from it, with a cargo rust/test/build and setting the `working-directory` as part of the gh action step value
4. Remove `choco install -y llvm` from windows build-chain, llvm is already installed on windows runners https://github.com/actions/runner-images/blob/d8d4d924ad21a8ba4931b2988f74bbb376cad021/images/win/toolsets/toolset-2022.json#L400
5. replace `nightly-2022-12-01` with `nightly` - fails with updated toolchain actions update, it looks like the underlying reason for it being pinned has been fixed https://github.com/rust-lang/rust/issues/105886
6. Skip upload release assets on the release workflow, where the tags don't correlate with a step to build artefacts.

### DIFF
--- a/.github/workflows/build-ffi.yml
+++ b/.github/workflows/build-ffi.yml
@@ -6,6 +6,7 @@ jobs:
   build:
     runs-on: ${{ matrix.operating-system }}
     strategy:
+      fail-fast: false
       matrix:
         operating-system: [ ubuntu-latest, windows-latest, macos-latest ]
         rust: [ stable ]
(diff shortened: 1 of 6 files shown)


rholshausen:
Nice one!

PR #308, opened 2023-08-07
fix(pact_verifier): Fix missing PATCH version in plugin's version

tienvx:
Fix https://github.com/pact-foundation/pact-plugins/issues/44

~~I don't know how to write test for this. Please show me how if it's required.~~ I guess the 'missing plugin' test is enough to cover this change?

### DIFF
--- a/rust/pact_ffi/tests/tests.rs
+++ b/rust/pact_ffi/tests/tests.rs
@@ -484,7 +484,7 @@ fn test_missing_plugin() {
   pactffi_verifier_shutdown(handle);
 
   expect!(result).to(be_equal_to(2));
-  expect!(output.to_string_lossy().contains("Verification execution failed: Plugin missing-csv:0.0 was not found")).to(be_true());
+  expect!(output.to_string_lossy().contains("Verification execution failed: Plugin missing-csv:0.0.3 was not found")).to(be_true());
 }
 
 // Issue #299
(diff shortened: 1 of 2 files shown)


rholshausen:
The reason I dropped the patch version, is because I had this problem (just making up the versions):
* Created a Pact test with plugin version 1.0.2
* Published the Pact file
* When I went to verify the Pact file, it failed due to a bug in the plugin
* Fixed the plugin a few times, it ended up at version 1.0.5
* Tried to verify the Pact file, but it was still failing because the Pact file was linked to version 1.0.2

At this point I realised that you probably always want the latest version, even with minor, not just patch.

tienvx:
Sorry I didn't know about this feature.

btw, I think this feature is not implemented for provider verification? Or at least it didn't print out the log that it's working e.g.

```
Version `0.0.3` is selected as latest version of `0.0` of csv plugin
```

If it's working, I still don't know what cause the bug https://github.com/pact-foundation/pact-plugins/issues/44

YOU54F:
> Sorry I didn't know about this feature.

That's okay, not your fault, I am not sure if we have documented, so will make a note of after we've resolved this

>btw, I think this feature is not implemented for provider verification? Or at least it didn't print out the log that it's working 

This may be the case, would need to check the code. I'll leave @uglyog to second check it

> If it's working, I still don't know what cause the bug https://github.com/pact-foundation/pact-plugins/issues/44

Once we confirm it is expected to be implemented and working, we can check retest but will consider that case when looking at this change.

Thanks Tien!

PR #312, opened 2023-08-14
chore: update 'could not load pacts' message to 'no pacts found'

bethesque:
What I'm trying to achieve is to change the message from "Could not load pacts from the pact broker" to "No pacts found matching the given consumer version selectors in pact broker" when there are no pacts returned in the pacts-for-verification response. The current message makes it sound like there was a problem during the actual fetching process. 

I don't think this is the right place, because this code seems to be handling an error, rather than dealing with a `pacts.length == 0` scenario, but I thought I'd start somewhere!

### DIFF
--- a/rust/pact_verifier/src/lib.rs
+++ b/rust/pact_verifier/src/lib.rs
@@ -1257,9 +1257,9 @@ async fn fetch_pact(
           buffer
         },
         Err(err) => {
-          error!("Could not load pacts from pact broker '{}': {}", broker_url, err);
+          error!("No pacts found in pact broker '{}': {}", broker_url, err);
           vec![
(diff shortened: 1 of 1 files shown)


rholshausen:
Maybe we should include the selectors that were used in the request?

bethesque:
Yes, that would be handy. I might need to learn a bit more rust to submit that though! Shall we start with this one and then improve it?

rholshausen:
It is going to need a refactor, as you mentioned, this code seems to be handling an error, rather than dealing with the source occurrence. I'm happy to take this on.

bethesque:
Thanks Ron. I'm confused where the message for pacts.length == 0 is coming from though, because I can't find any other instance of that text, and that's definitely the message people are getting when there are no pacts returned. Is there another codebase?

Issue #318, opened 2023-08-30
Multipart: Mismatch when comparing 2 identical strings


tienvx:
consumer's tests failed because of mismatch when comparing 2 identical strings (`testing` !== `testing`)

How to reproduce:
* `git clone https://github.com/tienvx/pact-php.git`
* `cd pact-php`
* `git checkout multipart-static-files`
* `composer install`
* `cd example/multipart/consumer/`
* `PACT_LOGLEVEL=trace phpunit`
* Expected: test passed without mismatches
* Actual: there are 2 mismatches when comparing 2 strings with matcher `ContentType(\"text/plain\")`
  * `pact_matching::matchers: String -> String: comparing 'testing' to 'testing' ==> false cascaded=false matcher=ContentType("text/plain")`
  * `pact_matching::matchers: String -> String: comparing 'Colten Ziemann' to 'Zoey Turcotte' ==> false cascaded=false matcher=ContentType("text/plain")`

Partial logs:

```
2023-08-30T04:42:58.653223Z DEBUG ThreadId(03) matches_with{self="testing" actual="testing" matcher=ContentType("text/plain") cascaded=false}: pact_matching::matchers: String -> String: comparing 'testing' to 'testing' ==> false cascaded=false matcher=ContentType("text/plain")
2023-08-30T04:42:58.653226Z TRACE ThreadId(03) pact_matching::matchers: match_values: &str -> &str result=Err(["Unable to match 'testing' using ContentType(\"text/plain\")"])
2023-08-30T04:42:58.653231Z DEBUG ThreadId(03) pact_matching::binary_utils: Comparing file part 'MimeFile { index: 2, name: "personal_note", content_type: Some("text/plain"), filename: "note.txt", data: b"testing", headers: {"content-disposition": "form-data; name=\"personal_note\"; filename=\"note.txt\"", "content-type": "text/plain"} }' to 'MimeFile { index: 2, name: "personal_note", content_type: Some("text/plain"), filename: "note.txt", data: b"testing", headers: {"content-disposition": "form-data; name=\"personal_note\"; filename=\"note.txt\"", "content-length": "7", "content-type": "text/plain"} }' at path '$.personal_note' -> BodyMismatches({"$": [BodyMismatch { path: "$", expected: Some(b"testing"), actual: Some(b"testing"), mismatch: "Unable to match 'testing' using ContentType(\"text/plain\")" }]})
```

```
[...]

PR #319, opened 2023-09-11
fix: Allow matching string values using content type matching rule

tienvx:
Fix https://github.com/pact-foundation/pact-reference/issues/318


mefellows:
I'm a bit confused about the use case here Tien, mind elaborating a bit?

If you want it exposed via FFI you'll need to expose the rule elsewhere also.

tienvx:
I explain it a different way in https://github.com/pact-foundation/pact-reference/issues/318#issuecomment-1713801862, hope it help.

PR #335, opened 2023-11-03
fix: Fallback to always generate a value to fix 'Could not generate a random TYPE from null'

tienvx:
Remove the errors similar to `Could not generate a random decimal from null`.

It then allow to use integration json like this:

```
{
    "pact:matcher:type": "null",
    "pact:generator:type": "RandomDecimal"
}
```

This change is for implementing compatibility suite in pact-php (and probably other languages that use FFI calls as well).

For more information, please take a look at this [discussion](https://pact-foundation.slack.com/archives/C02BXLDJ7JR/p1698980611664139)

### DIFF
--- a/rust/pact_models/src/generators/mod.rs
+++ b/rust/pact_models/src/generators/mod.rs
@@ -1000,40 +1000,18 @@ impl GenerateValue<Value> for Generator {
     matcher: &Box<dyn VariantMatcher + Send + Sync>
   ) -> anyhow::Result<Value> {
     debug!(context = ?context, "Generating value from {:?}", self);
+    let mut rnd = rand::thread_rng();
     let result = match self {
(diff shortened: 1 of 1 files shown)


rholshausen:
While this fixes the issue above, it regresses behavior of the random integer and decimal generators. 

What the previous code was doing was generating either a number or a string depending on the type of the value. With this change, it will always now generate a string value.

For instant, with input JSON:
```json
{
  "value": 100
}
```

It will now generate:

```json
{
  "value": "3498499"
}
```

and not 

```json
{
  "value": 3498499
}
```

I think the fix is to only change the fallback to generate either type of value.

tienvx:
I updated the code follow your suggestion

rholshausen:
Nice!

Issue #338, opened 2023-11-17
FFI calls > Support generators for async message


tienvx:
These scenarios in compatibility suite indicate that generators are supported for async message on consumer side:

- [Supports the use of generators with the message body](https://github.com/pact-foundation/pact-compatibility-suite/blob/main/features/V3/message_consumer.feature#L63-L69)
- [Supports the use of generators with message metadata](https://github.com/pact-foundation/pact-compatibility-suite/blob/main/features/V3/message_consumer.feature#L71C13-L78)

But currently I think generators are not working when using this `pactffi_message_reify` FFI call.

mefellows:
Reify doesn't support synchronous messages and a bunch of things. I think you're best using the message-related iterators to extract these values. The method should probably be marked as deprecated.

tienvx:
I don't know how to use those iterators methods yet. I can't find any example code. Look like there are no language using those methods yet?

btw thanks for the suggestion.

mefellows:
No worries!

Btw I use it in Go here: https://github.com/pact-foundation/pact-go/blob/master/internal/native/message_server.go#L397-L519.

tienvx:
I'm able to use those iterator methods, but there 2 small problems:
* The code is too long for me (15 lines of code vs 1 when using `pactffi_message_reify`)
* An annoying issue related to panic #344

I also updated `pactffi_message_reify` with this PR #342, so for now I will continue using it.

tienvx:
Closed as implemented

mefellows:
Nice one! I'm interested - does this work on synchronous messages also (both the request and/or response side)?

tienvx:
I didn't touch on synchronous message FFI calls. But I have a quick look at `synchronous_messages_iter` and I think it is already implemented:
* https://github.com/pact-foundation/pact-reference/blob/master/rust/pact_consumer/src/builders/message_iter.rs#L56
* https://github.com/pact-foundation/pact-reference/blob/master/rust/pact_consumer/src/builders/message_iter.rs#L76

Currently I can't implement synchronous message in compatibility suite in PHP so I can't test it.

mefellows:
oh yes, I know they support generators. I guess I was asking if you could use `pactffi_message_reify` on a sync message. I can look at code, I was being lazy :P

tienvx:
> if you could use pactffi_message_reify on a sync message

No, I don't think so.

PR #340, opened 2023-11-20
feat: Apply generators to message's contents

tienvx:
Apply generators to `pactffi_message_get_contents`

For #338

### DIFF
--- a/rust/pact_ffi/src/models/message.rs
+++ b/rust/pact_ffi/src/models/message.rs
@@ -26,6 +26,11 @@ use crate::models::pact_specification::PactSpecification;
 use crate::util::*;
 use crate::util::string::optional_str;
 
+use pact_matching::generators::generate_message;
+use pact_models::generators::GeneratorTestMode;
+use maplit::hashmap;
+use futures::executor::block_on;
+
 /*===============================================================================================
  * # Re-Exports
  *---------------------------------------------------------------------------------------------*/
@@ -160,6 +165,7 @@ ffi_fn! {
     /// a NULL message and a missing message body.
     fn pactffi_message_get_contents(message: *const Message) -> *const c_char {
         let message = as_ref!(message);
+        let message = block_on(generate_message(&message, &GeneratorTestMode::Consumer, &hashmap!{}, &vec![], &hashmap!{}));
 
         match message.contents {
             // If it's missing, return a null pointer.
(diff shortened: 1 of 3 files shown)


rust/pact_ffi/src/models/message.rs
```diff
@@ -160,6 +165,7 @@ ffi_fn! {
     /// a NULL message and a missing message body.
     fn pactffi_message_get_contents(message: *const Message) -> *const c_char {
         let message = as_ref!(message);
+        let message = block_on(generate_message(&message, &GeneratorTestMode::Consumer, &hashmap!{}, &vec![], &hashmap!{}));
```
rholshausen, line 168: `block_on` will block the current thread until the future has run to completion. If this function is called concurrently (i.e. from two different threads from the other side of the FFI at the same time), I think it will cause a deadlock.

rholshausen: No, I think this is ok as long as it is only called via FFI. If it is called from another FFI function that uses a Tokio runtime, it will block the runtime and cause a deadlock.

tienvx: Currently I'm only interested in getting generators work with `pactffi_message_reify`.

Should I close PRs for `pactffi_message_get_contents` (this PR) and `pactffi_message_metadata_iter_next` (#341) to lower the risk?

rholshausen: No, I think it is good for now. As long as these are only called via FFI. I can't think of a reason why another FFI function would call this. They should call the underlying functions (i.e. `generate_message`) themselves.

rholshausen: Maybe just add a line to the doc comment under the Safety section that the function must only be called via foreign languages and not Rust code, otherwise a deadlock could result.

tienvx: Can you help me add a commit to this PR? I will copy the comment to other methods

rholshausen: The commits are in your repo, which I won't be able to do. Here is the comment:

```rust
/// # Safety
///
/// The returned string must be deleted with `pactffi_string_delete` and can outlive the message.
/// This function must only ever be called from a foreign language. Calling it from a Rust function
/// that has a Tokio runtime in its call stack can result in a deadlock.
```

tienvx: Thanks @rholshausen

Issue #337, opened 2023-11-17
Strange behavior of `Values` matcher


tienvx:
This is the description of `Values` matcher from [specification v4](https://github.com/pact-foundation/pact-specification/tree/version-4#supported-matching-rules):

> Match the values in a map, ignoring the keys

But from what I tested, I think it doesn't ignore the keys, and the way it "Match the values in a map" doesn't make sense to me.

tienvx:
Here are some behaviors on array:

1. Array
    1. All values present:
        * Consumer: `Values(["a", "bb", "ccc"])`
        * Provider: `["a", "bb", "ccc"]`
        * Result: OK
    1. Missing value at the **beginning**:
        * Consumer: `Values(["a", "bb", "ccc"])`
        * Provider: `["bb", "ccc"]`
        * Result: NOT OK
        * Output:

                ```
                $.values[1] -> Expected 'ccc' (String) to be equal to 'bb' (String)
                $.values[0] -> Expected 'bb' (String) to be equal to 'a' (String)
                ```

        * My comment: output make sense to me
    1. Missing value at the **end**:
        * Consumer: `Values(["a", "bb", "ccc"])`
        * Provider: `["a", "bb"]`
        * Result: OK
        * My comment: it's weird, I don't understand why it accept missing value (and doesn't accept additional value. Postel's law but in revert?)
    1. Duplicated first value at the **end**:
        * Consumer: `Values(["a", "bb", "ccc"])`
        * Provider: `["a", "bb", "ccc", "a", "a", "a"]`
        * Result: OK
    1. Duplicated other values at the **end**:
        * Consumer: `Values(["a", "bb", "ccc"])`
        * Provider: `["a", "bb", "ccc", "bb", "ccc"]`
        * Result: NOT OK
        * Output:

                ```
               $.values[3] -> Expected 'bb' (String) to be equal to 'a' (String)
               $.values[4] -> Expected 'ccc' (String) to be equal to 'a' (String)
                ```

        * My comment: output **doesn't** make sense to me
    1. Additional value at the **beginning**:
        * Consumer: `Values(["a", "bb", "ccc"])`
        * Provider: `["dddd", "a", "bb", "ccc"]`
        * Result: NOT OK
        * Output:

                ```
                $.values[3] -> Expected 'ccc' (String) to be equal to 'a' (String)
                $.values[0] -> Expected 'dddd' (String) to be equal to 'a' (String)
[...]

PR #365, opened 2024-01-22
test(pact_matching): Test Values matcher with json body

tienvx:
These tests describe weird behavior of `Values` matcher mentioned in #337

* The `Values` matcher is mentioned 2 times in the [specification](https://github.com/pact-foundation/pact-specification/tree/version-4#supported-matching-rules)
  * About `Values` matcher itself: `Match the values in a map, ignoring the keys`
  * About `EachValue` matcher: `... For maps, delgates to the Values matcher.`
* I have no clue how it should work, except it `ignoring the keys`
* I take a look at the code
  * The code handling matching for `maps` and `lists` is a bit complicated.
  * But I can see the `first value` play an important role. It explain why `missing values` is ok, `duplicated first value` is ok
  * Look like the code for `Values` matcher is empty, so it's not implemented, but rely on a general code that handle matching for `maps` and `lists`? I'm not sure.
  * It doesn't `ignore the keys` at the moment.

So I came up with my own "rules" for this matcher (already in the issue above):

For **lists**/json array/array without keys:

1. All values must be present
2. The order of values is not important
3. Missing values are NOT OK
4. Additional values are NOT OK
5. Duplicated values may be OK?

For **maps**/json object/array with keys:

1. Keys must be ignored
2. Other than that, same with lists

 **I don't know how to implement these rules in the code yet when they are accepted.**

Compare with other matchers:

| Behavior   |      Values      | ArrayContains | EachValue |
|----------|:-------------:|------:|------:|
| ignore keys | yes | yes | yes |
| support value's matching rules |  no | yes | yes |
| value's order | not important | not important | not important |
| missing values |not ok | not ok | not sure the behavior when actual values is an empty list/array, not ok I guess |
| additional values |not ok | ok | not ok |
| duplicated values |ok | ok | ok |

### DIFF
--- a/rust/pact_matching/src/json.rs
+++ b/rust/pact_matching/src/json.rs
@@ -1279,3 +1279,91 @@ mod tests {
     expect!(result).to(be_err());
   }
 }
+
+#[cfg(test)]
+mod tests2 {
+  use expectest::prelude::*;
+  use maplit::hashmap;
+  use rstest::rstest;
+  use serde_json::{json, Value};
+
+  use pact_models::matchingrules_list;
+  use pact_models::matchingrules::MatchingRule;
+  use crate::{CoreMatchingContext, DiffConfig};
+
+  use super::*;
+
+  #[rstest]
+  //                                                    config,                          actual_json,                                           is_ok
+  #[case::no_unexpected_keys_same_values(               DiffConfig::NoUnexpectedKeys,    json!({"a": "a", "b": "bb", "c": "ccc"}),              true)]
+  #[case::no_unexpected_keys_missing_first_value(       DiffConfig::NoUnexpectedKeys,    json!({"b": "bb", "c": "ccc"}),                        true)] // should be err
+  #[case::no_unexpected_keys_missing_last_value(        DiffConfig::NoUnexpectedKeys,    json!({"a": "a", "b": "bb"}),                          true)] // should be err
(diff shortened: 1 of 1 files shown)


rholshausen:
This is not true:
> The order of values is not important
How can you ensure this? Basically, the Values matcher converts a map into a list of values, and then lets the normal matching logic continue as if a Type matcher was specified. If the order of the values changes, the matching will fail if there are different types. For anything other than a map, it just acts like a Type matcher.

This matcher has been superseded by the EachValue and EachKey matchers, as the logic to implement it was very messy and those two matchers are much clearer on what they do.

tienvx:
> How can you ensure this? Basically, the Values matcher converts a map into a list of values, and then lets the normal matching logic continue as if a Type matcher was specified. If the order of the values changes, the matching will fail if there are different types. For anything other than a map, it just acts like a Type matcher.

Is it really `Type` matcher? I think it's `Equality` matcher

tienvx:
I'm still confused about `Values` matcher.

> This matcher has been superseded by the EachValue and EachKey matchers, as the logic to implement it was very messy and those two matchers are much clearer on what they do.

Should I simply just remove the `Values` matcher from `pact-php` so we don't allow user to use it, so we don't have to maintain it?

cc @mefellows , @YOU54F

tienvx:
Look like `pact-js` has a similar approach recently: deprecated `eachKeyLike` (really `values` matcher) and replace it with `eachKey` and `eachValue`

PR #380, opened 2024-02-06
Add matching rules

JP-Ellis:
## Summary

This ultimately covers two things:

1. Adding a `merge` method to the `MatchingRules` model; and,
2. Adding a `pactffi_with_matching_rules` FFI function.

In order to simplify development, the `pact_models` dependency uses a relative path so the latest local code is used during development. The `patch` section in the `Cargo.toml` file is used to ensures this is done also transitive dependencies that originate from the workspace.

## Commits

* relase(pact_models): v1.2.0 (f3b1f6f9) <JP-Ellis>

  Signed-off-by: JP-Ellis <josh@jpellis.me>

* release(pact_ffi): v0.5.0 (e5d87a1a) <JP-Ellis>

  Signed-off-by: JP-Ellis <josh@jpellis.me>

* feat(ffi): add with_matching_rules (6bb6392d) <JP-Ellis>

  While it is already possible to add matching rules through the various other
  methods, it was not possible to add a matching rule directly. This adds a new
  `with_matching_rules` FFI function.

  This was inspired by the implementation of the V2 compatibility suite in Pact
  Python.

  Signed-off-by: JP-Ellis <josh@jpellis.me>

* feat(models): add merge method (355cde2f) <JP-Ellis>

  Add a `merge` method on MatchingRules allowing MatchingRules to be combined
  together. This is intended to ultimately be used by a new
  `with_matching_rules` FFI function.

  Signed-off-by: JP-Ellis <josh@jpellis.me>

* chore: use local pact_models (155d1221) <JP-Ellis>

  The `pact_models` dependency underpins a number of the other packages within
  the workspace, and the `pact-plugin-driver` package external to this
  workspace. This committee ensures the local `pact_models` library is
  consistently used during development by:

  1. Using the `path = "..."` for packages within the workspace; and, 2. Using a
  `patch` in `Cargo.toml` to override transitive
    dependencies (such as the one that comes from `pact-plugin-driver`).

  Signed-off-by: JP-Ellis <josh@jpellis.me>

* chore(rust): simplify workspace inclusions (28874261) <JP-Ellis>

  Signed-off-by: JP-Ellis <josh@jpellis.me>


rust/pact_ffi/src/mock_server/handles.rs
```diff
@@ -1809,6 +1809,81 @@ pub extern fn pactffi_with_binary_file(
...
+      Ok(rules) => rules,
+      Err(err) => {
+        error!("with_matching_rules: Failed to load the matching rules: {}", err);
+        return Ok(false);
+      }
+    };
+
+    interaction.with_interaction(&move |_, _, inner| {
+      if let Some(reqres) = inner.as_v4_http_mut() {
+        reqres.request.matching_rules_mut().merge(&rules);
```
rholshausen, line 1867: This function only allows setting the matching rules for the request. It needs to also allow setting the matching rules for the response.

rholshausen (changes requested):
I don't think we need the version bump on the FFI lib, as it is only adding a new FFI function. This will not break existing use.

JP-Ellis:
Closing as it has been cherry-picked into main

PR #381, opened 2024-02-06
feat: Print provider name in 'no pacts found' message to help debugging

tienvx:
Provider name is optional, and default value is 'provider'.

When verifier is configured without provider name (probably new user not knowing it, or forget about it), and there is (high chance that) no provider has the name 'provider', verifier print this message:

`No pacts found matching the given consumer version selectors in pact broker`

This message is not wrong, but it is hard to debug in this situation


rholshausen:
Good one!

PR #384, opened 2024-02-09
Support metadata rules for sync message

stan-is-hate:
This PR fixes an issue where metadata matching rules were not stored in the pact file. 
First time writing any rust code, please let me know if something is not idiomatic or optimal.


rholshausen:
All, good. Thanks for adding all the additional tests.

PR #391, opened 2024-02-26
Chore/fix pactffi release

JP-Ellis:
## Summary

Hopefully fixes the issue with the release pipeline discovered when creating `0.4.18`.

The main changes:

- A refactor of the `release-osx.sh` script (primarily DRYing code), and importantly **removes the setting of `SDKROOT` and `MAXOSX_DEPLOYMENT_TARGET`**. @YOU54F, I believe you added these, so please let me know if that is likely to cause an issue.
- A refactor of the release pipeline (primarily DRYing code)
- Updating to using the latest OS runners

## Motivation

Hopefully, to resolve the release script failure.

I was thinking of updating the release script to run on PRs as well (to detect failures, though without the creation of artefacts), but ultimately opted not to do so as the pipeline can be slow and may not be necessary if the other pipeline succeed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,88 +2,142 @@ name: Release workflow
 
 on:
   release:
-    types: [published]
+    types:
+      - published
+
+  pull_request:
+    branches:
+      - master
+
+concurrency:
+  group: release-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
   build-release:
     runs-on: ${{ matrix.operating-system }}
     strategy:
       matrix:
-        operating-system: [ubuntu-20.04, windows-latest, macos-11]
+        include:
+          - operating-system: ubuntu-20.04
+            targets: x86_64-unknown-linux-gnu,x86_64-unknown-linux-musl,aarch64-unknown-linux-gnu,aarch64-unknown-linux-musl
+          - operating-system: windows-2019
+            targets: aarch64-pc-windows-msvc,x86_64-pc-windows-msvc
+          - operating-system: macos-12
+            targets: aarch64-apple-darwin,x86_64-apple-darwin
+      fail-fast: false
+
     env:
       pact_do_not_track: true
+
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
+
       - name: Install stable Rust toolchain
-        if: runner.os == 'Linux'
         uses: dtolnay/rust-toolchain@stable
         with:
           toolchain: stable
-          targets: aarch64-unknown-linux-gnu,x86_64-unknown-linux-musl
-      - name: Install stable Rust toolchain
-        if: runner.os == 'Windows'
-        uses: dtolnay/rust-toolchain@stable
+          targets: ${{ matrix.targets }}
+
+      - name: Rust caching
+        uses: Swatinem/rust-cache@v2
         with:
-          toolchain: stable
-          targets: aarch64-pc-windows-msvc,x86_64-pc-windows-msvc
-      - name: Install stable Rust toolchain
-        if: runner.os == 'MacOS'
-        uses: dtolnay/rust-toolchain@stable
-        with:
-          toolchain: stable
-          targets: aarch64-apple-darwin,x86_64-apple-darwin
-      - uses: maxim-lobanov/setup-xcode@v1
-        with:
-          xcode-version: 12.4
-        if: runner.os == 'MacOS'
+          workspaces: rust
+
       - name: Set up QEMU
         if: runner.os == 'Linux'
         uses: docker/setup-qemu-action@v3
+
       - name: Set up Docker Buildx
         if: runner.os == 'Linux'
         uses: docker/setup-buildx-action@v3
-      - if: startsWith(github.ref, 'refs/tags/pact_mock_server_cli') && runner.os == 'Linux'
-        run: ./release-linux.sh
+
+      - name: Platform abbreviation
+        id: platform-abbreviation
+        shell: bash
+        run: |
+          if [[ "${{ runner.os }}" = "Linux" ]]; then
+            echo "platform=linux" >> "$GITHUB_OUTPUT"
+          elif [[ "${{ runner.os }}" = "Windows" ]]; then
+            echo "platform=win" >> "$GITHUB_OUTPUT"
+          elif [[ "${{ runner.os }}" = "macOS" ]]; then
+            echo "platform=osx" >> "$GITHUB_OUTPUT"
+          else
+            echo "Unknown platform"
+            exit 1
+          fi
+
+      - name: Cargo flags
+        id: cargo-flags
+        shell: bash
+        run: |
+          if [[ "${{ github.event_name }}" = "release" ]]; then
+            echo "flags=--release" >> "$GITHUB_OUTPUT"
+          else
+            echo "flags=" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Build mock server CLI
+        if: |
+          github.event_name == 'push' ||
+          github.event_name == 'pull_request' ||
+          startsWith(github.ref, 'refs/tags/pact_mock_server_cli')
+        shell: bash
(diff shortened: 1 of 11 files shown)


mefellows:
It looks like I added those lines (following git blame). Looking at the history, might have been an M1 thing: https://github.com/pact-foundation/pact-reference/blame/28eb5c2e0c4b905490a92487fdc925f14e6eb145/rust/pact_ffi/release-osx.sh

I assume it’s not related to the `install_name` shenanigans. Perhaps it was an attempt to build on an older version so that it would be compatible with older OSX versions?

YOU54F:
we should have it set but to the oldest deployment target we support. macos arm64 only from big sur.if we update runners we will build against a newer xcode and the binary wont be compatible with older mac versionsOn 26 Feb 2024, at 10:19, Matt Fellows ***@***.***> wrote:﻿
It looks like I added those lines (following git blame). Looking at the history, might have been an M1 thing: https://github.com/pact-foundation/pact-reference/blame/28eb5c2e0c4b905490a92487fdc925f14e6eb145/rust/pact_ffi/release-osx.sh
I assume it’s not related to the install_name shenanigans. Perhaps it was an attempt to build on an older version so that it would be compatible with older OSX versions?

YOU54F:
install_name issue was a totaly diff apple introduced bug

YOU54F:
I did an action run to echo out the values set after running the xcode setup action

https://github.com/YOU54F/pact-reference/actions/runs/8028038268/job/21932742579

```
swiftc -version
xcrun --show-sdk-version
xcrun --show-sdk-path
xcrun --show-sdk-platform-version
export SDKROOT=$(xcrun -sdk macosx11.1 --show-sdk-path)
export MACOSX_DEPLOYMENT_TARGET=$(xcrun -sdk macosx11.1 --show-sdk-platform-version)
echo $SDKROOT
echo $MACOSX_DEPLOYMENT_TARGET
```

With xcode 12.4

```
Apple Swift version 5.3.2 (swiftlang-1200.0.45 clang-1200.0.32.28)
Target: x86_64-apple-darwin20.6.0
11.1
/Applications/Xcode_12.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk
11
/Applications/Xcode_12.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk
11.1
```

With the default xcode version in the image

```
12.1
/Applications/Xcode_13.2.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk
12.1
```

Setting MACOSX_DEPLOYMENT_TARGET to 11.1 is sensible, this targets big sur and later. We should just be able to set that for both targets, this would ensure our binaries work with big sur and later.

https://en.wikipedia.org/wiki/MacOS_version_history

Note, I use `export MACOSX_DEPLOYMENT_TARGET=12.2` in traveling-ruby and that works as far back as Catalina (Catalina being intel only)

https://github.com/YOU54F/traveling-ruby/blob/traveling-pact/osx/internal/reset_environment.sh#L23

YOU54F:
>I was thinking of updating the release script to run on PRs as well (to detect failures, though without the creation of artefacts), but ultimately opted not to do so as the pipeline can be slow and may not be necessary if the other pipeline succeed.

I think we should test the builds outside of releases, and maybe smoke test them.

Ideally I'd like to split the runs by indiv targets as it makes it much easier to work out which is broken

> Updating to using the latest OS runners

I would hold off with this, updating the runners has implications on anything built with cargo, over cross due to linked system libs

.github/workflows/release.yml
```diff
@@ -9,77 +9,93 @@ jobs:
     runs-on: ${{ matrix.operating-system }}
     strategy:
       matrix:
-        operating-system: [ubuntu-20.04, windows-latest, macos-11]
+        include:
+          - operating-system: ubuntu-latest
```
rholshausen, line 13: We shouldn't run latest (apart from Windows) as GitHub could update those and that can cause issues.

JP-Ellis: Why not for Windows? `windows-latest` currently points to `windows-2022` and the other version available is `windows-2019`.

rholshausen: Windows Win32API is very stable. It doesn't seem to matter what version you build on, it will run on older versions. But if you want to be consistent ...

JP-Ellis:
@YOU54F Thanks for explaining why we should target the 11.1 SDK. Is there a reason why this is only done for the `aarch64` build and not the `x86_64` build?

JP-Ellis:
And while updating this PR to resolve the conflicts introduced by installing XCode, I found that Apple dropped support for macOS 11 (Big Sur) 5 months ago: https://endoflife.date/macos

Should we really be targeting versions which are no longer supported?

YOU54F:
> @YOU54F Thanks for explaining why we should target the 11.1 SDK. Is there a reason why this is only done for the `aarch64` build and not the `x86_64` build?

I would imagine it was just being explicitly set to the first sdk that arm64 supported, it would probably make sense to set it explicitly for both so we can be consistent (similar to using a fixed glibc/musl version on linux), and it would mean that updating to a new runner isn't an issue. technically we could set the deployment targets and switch to a macos-14 runner to get that sweet arm64 powaaaah and still get equivalant builds out 

we _may_ want to target earlier deployment targets for x86_64 builds, as there are still hold-outs on older versions, but I haven't seen reports of anyone on old versions of macos with intel processors with issues, so it might be a moot point

also if you want to get really geeky, its possible with the right macos sdk, to build 32bit binaries on an arm64 machine (albeit with rosetta) https://github.com/mrpippy/XcodeNueve

YOU54F:
> And while updating this PR to resolve the conflicts introduced by installing XCode, I found that Apple dropped support for macOS 11 (Big Sur) 5 months ago: https://endoflife.date/macos

yeah macos-11 runners will be going removed from gh soon

https://github.com/actions/runner-images/issues/9255

> Should we really be targeting versions which are no longer supported?

We shouldn't really be supporting them but we should build for the earlier possible we can, as long as they are forward compatible. not everyone is able to upgrade to the latest and greatest, especially with regards to customers who are using Pact to protect not just bleeding edge software, but heritage apps, on heritage servers alike xD

JP-Ellis:
Thanks @YOU54F 

I've updated the PR with the following changes:

- Use `macos-12`, in anticipation of `macos-11` being removed soon
- Target SDK `12.x`, since `11` is now unmaintained.

I'll test out it first in a fork to make sure the changes work. With the upgrade to SDK 12, it may no longer be necessary to install XCode separately (which I will test).

rholshausen:
> Should we really be targeting versions which are no longer supported?
We should be targeting versions that our users are using, not what Apple or Oracle or Microsoft tell us to.

YOU54F:
> Thanks @YOU54F
> 
> I've updated the PR with the following changes:
> 
> * Use `macos-12`, in anticipation of `macos-11` being removed soon
> * Target SDK `12.x`, since `11` is now unmaintained.
> 
> I'll test out it first in a fork to make sure the changes work. With the upgrade to SDK 12, it may no longer be necessary to install XCode separately (which I will test).

Cheers dude, I'd maybe consider trying out an macos-14 build for a speed comparison.

Testing is a pain, unless you have a suite of vm's that go way back for intel, and there are no big sur arm64 vm's we can't test out if the binaries work that far back. I happen to have vm's back to high sierra on my intel hackintosh because I was testing the traveling-ruby builds, so can give them a whirl

I'm not sure you even need the full xcode build, the command line tools along will probably suffice. If it builds with it hard-coded version of 11.1 without having to install xcode expliclity with the action (which is probably just setting up a missing symlink as its apparently installed on the box anyway!), and it builds it'll probably be grand.

JP-Ellis:
I've just push the most recent reefactor. I've refactored all of the scripts (for all platforms), and it will now run on PRs too.

I've also enabled some caching to hopefully bring down the build time.

PR #401, opened 2024-03-14
feat(ffi): add `add_text_comment`

JP-Ellis:
While the addition of `set_comment` allows for the comments to be used as per the Pact specification, it did not follow the original intent for comments (see the discussion in pact-foundation/pact-compatibility-suite#8).

The new `add_text_comment` function allows comments to be added repeatedly and appended into an array as intended in pact-foundation/pact-specification#45.

### DIFF
--- a/rust/pact_ffi/src/mock_server/handles.rs
+++ b/rust/pact_ffi/src/mock_server/handles.rs
@@ -2171,6 +2171,10 @@ ffi_fn!{
   /// Note that a `value` that deserialize to a JSON null will result in a
   /// comment being added, with the value being the JSON null.
   ///
+  /// Note that the `text` key is special and is used by
+  /// [`pactffi_add_text_comment`] to append comments to the array of comments.
+  /// Overwriting the `text` key is allowed, but should be done with caution.
+  ///
   /// # Safety
   ///
   /// The comments parameter must be a valid pointer to a NULL terminated UTF-8,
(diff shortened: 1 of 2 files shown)


rust/pact_ffi/src/mock_server/handles.rs
```diff
@@ -2227,6 +2231,66 @@ ffi_fn!{
...
+      None => {
+        error!("add_text_comment: Comment value is not valid (NULL or non-UTF-8)");
+        return Err(anyhow!("Comment value is not valid (NULL or non-UTF-8)"));
+      }
+    };
+
+    let ensure_append = |comments: &mut serde_json::Value| {
+      match comments {
+        serde_json::Value::Array(_) => {
+          comments.as_array_mut().unwrap().push(serde_json::Value::String(comment.to_string()));
```
rholshausen, line 2266: You've already pattern matched on the array, you can just use the inner Vec and not need to call as_array and unwrap

rust/pact_ffi/src/mock_server/handles.rs
```diff
@@ -2227,6 +2231,66 @@ ffi_fn!{
...
+        },
+        _ => {
+          *comments = serde_json::Value::Array(vec![serde_json::Value::String(comment.to_string())]);
+        }
+      }
+    };
+
+    interaction.with_interaction(&|_, _, inner| {
+      if let Some(reqres) = inner.as_v4_http_mut() {
```
rholshausen, line 2275: This feels unnecessary. Comments are independent of the interaction type.

rholshausen: Yes, the `V4Interaction` trait has `fn comments_mut(&mut self)` function.

rholshausen (changes requested):

JP-Ellis:
Thanks, I missed the `comments_mut()` method. I've also taken the opportunity to refactor the previous `set_comment` method (as I did the unnecessary matching there too).

And good catch with the matching 😅

PR #403, opened 2024-03-16
chore(ci): verifier - fix ref to build_aarch64 func post building wit…

YOU54F:
…h musl only

Thanks @rholshausen for building our executables with musl for both architectures on linux, dropping the need to retain two sets of binaries for linux 

Failure seen here

https://github.com/pact-foundation/pact-reference/actions/runs/8291686804/job/22691789837


rholshausen:
Thank you, Sir!

Issue #399, opened 2024-03-06
Support composing matching rules in FFI


mefellows:
Pact JVM supports combining matching rules using the `AND` logical expression.

It is currently not possible to do this via the FFI (e.g. using https://docs.pact.io/implementation_guides/rust/pact_ffi/integrationjson as a guide).

The matching definition expressions do support nesting, however they are not supported via the FFI in the integration JSON.

There are two options to support this:

1. update the FFI to allow multiple matching rules
2. update the FFI to support the matching definition expressions

Ideally, a method that allows the most flexible

**Example**

From a request in [slack](https://pact-foundation.slack.com/archives/C9VBGLUM9/p1709303346986579).

How might you write a matcher to check both the keys and values and the same time (use `eachKeyMatches` and `eachValueMatches` simultaneously):

```ts
    enum Keys {
      One = "One",
      Two = "Two",
    }

    interface Value {
      min: number
      max: number
    }

    type R = Record<Keys, Value>
    const record: R  = {
      [Keys.One]:
        {
          min: 1,
          max: 2
        }
    }

  it("check record", async () => {
    messagePact
      .expectsToReceive("record")
      .withContent({
        // What need to write here to check type R ?
      })
  })
```

JP-Ellis:
The work required for this should be partially covered by #383. I had to implement that to allow low-level JSON matching rules to be given straight to the FFI, because I found myself rewriting a matching rule parser in Python.

The FFI is the low-level implementation which will take the bare matching rule JSON, and the downstream libraries would likely need to implement a high level abstraction to help build these matching rules in a manner which makes most sense to each language.

mefellows:
Nice! 

> and the downstream libraries would likely need to implement a high level abstraction to help build these matching rules in a manner which makes most sense to each language.

Yes, and we should probably update the compatibility suite also if that's not already covered.

JP-Ellis:
Well, I added this _because_ of the compatibility suite. I found myself implementing a parser for the matching rules to convert them into the other JSON format as described in [integrationJson](https://docs.pact.io/implementation_guides/rust/pact_ffi/integrationjson).

This in itself begs the questions as to why we even had these two different formats, though I am sure there's a good reason.

sergewar:
Hi. I see that issue closed but maybe can you provide the code example how to use this new feature?

`// What need to write here to check type R ?`

rholshausen:
This has been implemented in the Rust core, but still needs work to use it in Pact-JS

[...]

PR #405, opened 2024-03-20
Feat/composing matching rules

rholshausen:
Implements #399 

TLDR - This allows an array to be specified with the `pact:matcher:type` attribute to be able to have multiple matching rules.

For example, you could do something like (just making this up, may not be valid):

```json
{
    "pact:matcher:type": [
        { "pact:matcher:type": "regex", "regex": "\\w{3}-\\d+" },
        { "pact:matcher:type": "include", "value": "XYZ" }
    ]
}
```

### DIFF
--- a/rust/pact_ffi/IntegrationJson.md
+++ b/rust/pact_ffi/IntegrationJson.md
@@ -157,3 +157,86 @@ Here the `interests` attribute would be expanded to
   }
 }
 ```
+
+## Supporting multiple matching rules
+
+Matching rules can be combined. These rules will be evaluated with an AND (i.e. all the rules must match successfully 
+for the result to be successful). The main reason to do this is to combine the `EachKey` and `EachValue` matching rules
+on a map structure, but other rules make sense to combine (like the `include` matcher).
+
+To provide multiple matchers, you need to provide an array format. 
+
+For example, assume you have an API that returns results for a document store where the documents are keyed based on some index:
+```json
+{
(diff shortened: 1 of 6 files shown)


rholshausen:
As I was in that space, I also added support for the matching rule definition expressions. So you also do

```json
{
    "pact:matcher:type": "eachKey(matching(regex, '\\w{3}-\\d+', 'AUK-155332')), eachValue(matching(type, ''))",
}
```

Just note that using references (`matching($'Example')`)  will not work.

JP-Ellis (approved):
Looks all pretty sensible to me.

Should the compatibility suite also be updated to make sure downstream libraries consistently use these?

rholshausen:
> Looks all pretty sensible to me.
> 
> Should the compatibility suite also be updated to make sure downstream libraries consistently use these?

This is more an implementation detail. The language implementations can choose to use this form or use the FFI functions directly. The compatibility suite should assert the behavior, regardless.

mefellows:
> > Looks all pretty sensible to me.
> > Should the compatibility suite also be updated to make sure downstream libraries consistently use these?
> 
> This is more an implementation detail. The language implementations can choose to use this form or use the FFI functions directly. The compatibility suite should assert the behavior, regardless.

Yeah it was more the point of being able to add multiple matchers to a single field.

Ditto expressions.

This looks good though!

mefellows (approved):

PR #410, opened 2024-04-17
Parse user-agent as single value headers

kageru:
User agents can and do contain commas, so they shouldn’t be split. Included is a test using the UA of my current browser.


rholshausen:
Thank you for the PR

PR #411, opened 2024-04-22
Feat/optional query parameter values

rholshausen:
Supports optional query parameter values. With this change we can support:
* `name=value` - query parameter with a value
* `name=` - query parameter with empty value
* `name` - query parameter with no value

### DIFF
--- a/compatibility-suite/tests/v3_steps/generators.rs
+++ b/compatibility-suite/tests/v3_steps/generators.rs
@@ -257,7 +257,10 @@ fn the_request_will_match(
     let parameter = key_regex.captures(request_part.as_str()).unwrap().get(1).unwrap().as_str();
     if let Some(query) = &world.generated_request.query {
       if let Some(value) = query.get(parameter) {
-        if value.iter().all(|v| regex.is_match(v.as_ref())) {
+        if value.iter().all(|v| {
+          let v = v.as_ref().map(|v| v.as_str()).unwrap_or_default();
+          regex.is_match(v)
+        }) {
           Ok(())
         } else {
           Err(anyhow!("Request query parameter {} has a value that does not match {}", parameter, regex))
--- a/rust/pact_ffi/src/mock_server/handles.rs
+++ b/rust/pact_ffi/src/mock_server/handles.rs
@@ -840,7 +840,21 @@ pub extern fn pactffi_with_query_parameter(
 /// pactffi_with_query_parameter_v2(handle, "id", 0, value);
 /// ```
 ///
-/// For a query parameter with no value, the value parameter can be set to a NULL pointer.
+/// For query parameters with no value, two distinct formats are provided:
+///
+/// 1. Parameters with blank values, as specified by `?foo=&bar=`, require an empty string:
+///
+/// ```c
+/// pactffi_with_query_parameter_v2(handle, "foo", 0, "");
+/// pactffi_with_query_parameter_v2(handle, "bar", 0, "");
+/// ```
+///
+/// 2. Parameters with no associated value, as specified by `?foo&bar`, require a NULL pointer:
+///
+/// ```c
+/// pactffi_with_query_parameter_v2(handle, "foo", 0, NULL);
+/// pactffi_with_query_parameter_v2(handle, "bar", 0, NULL);
+/// ```
 ///
 /// # Safety
 /// The name parameter must be a valid pointer to a NULL terminated string. If the value
--- a/rust/pact_ffi/src/mock_server/handles.rs
+++ b/rust/pact_ffi/src/mock_server/handles.rs
@@ -771,11 +771,11 @@ pub extern fn pactffi_with_query_parameter(
             if index >= values.len() {
               values.resize_with(index + 1, Default::default);
             }
-            values[index] = value;
+            values[index] = Some(value);
           } else {
-            let mut values: Vec<String> = Vec::new();
+            let mut values: Vec<Option<String>> = Vec::new();
             values.resize_with(index + 1, Default::default);
-            values[index] = value;
+            values[index] = Some(value);
             q.insert(name.to_string(), values);
           };
           q
@@ -784,9 +784,9 @@ pub extern fn pactffi_with_query_parameter(
           path.push_field(name).push_index(index);
           #[allow(deprecated)]
           let value = from_integration_json(&mut reqres.request.matching_rules, &mut reqres.request.generators, &value.to_string(), path, "query");
-          let mut values: Vec<String> = Vec::new();
+          let mut values: Vec<Option<String>> = Vec::new();
           values.resize_with(index + 1, Default::default);
-          values[index] = value;
+          values[index] = Some(value);
           Some(hashmap! { name.to_string() => values })
         });
         !mock_server_started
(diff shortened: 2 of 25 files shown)


rust/pact_ffi/src/mock_server/handles.rs
```diff
@@ -839,8 +839,12 @@ pub extern fn pactffi_with_query_parameter(
 /// const char* value = "{\"value\":[\"2\"], \"pact:matcher:type\":\"regex\", \"regex\":\"\\\\d+\"}";
 /// pactffi_with_query_parameter_v2(handle, "id", 0, value);
 /// ```
+///
+/// For a query parameter with no value, the value parameter can be set to a NULL pointer.
+///
 /// # Safety
-/// The name and value parameters must be valid pointers to NULL terminated strings.
+/// The name parameter must be a valid pointer to a NULL terminated string. If the value
+/// parameter is not NULL, it must point to a valid NULL terminated string.
```
JP-Ellis, line 847: I think it would be good to make the documentation more explicit, with perhaps some example usage to distinguish an "empty value" from a "null".

E.g.:

---

For query parameters with no value, two distinct formats are provided:

1. Parameters with blank values, as specified by `?foo=&bar=`, require an empty string:

   ```c
   pactffi_with_query_parameter_v2(handle, "foo", 0, "");
   pactffi_with_query_parameter_v2(handle, "bar", 0, "");
   ```

2. Parameters with no associated valued, as specified by `?foo&bar`, require a `NULL` pointer:

   ```c
   pactffi_with_query_parameter_v2(handle, "foo", 0, NULL);
   pactffi_with_query_parameter_v2(handle, "bar", 0, NULL);
   ```

Note that the two formats are distinct; however, support for either variant can be provided through the integration JSON:

```c
const char* value = "...";
pactffi_with_query_parameter_v2(handle, "foo", 0, value);
```

compatibility-suite/tests/v3_steps/generators.rs
```diff
@@ -257,7 +257,10 @@ fn the_request_will_match(
     let parameter = key_regex.captures(request_part.as_str()).unwrap().get(1).unwrap().as_str();
     if let Some(query) = &world.generated_request.query {
       if let Some(value) = query.get(parameter) {
-        if value.iter().all(|v| regex.is_match(v.as_ref())) {
+        if value.iter().all(|v| {
+          let v = v.as_ref().map(|v| v.as_str()).unwrap_or_default();
+          regex.is_match(v)
```
JP-Ellis, line 262: Is the loss of distinction intentional?

rust/pact_matching/src/lib.rs
```diff
@@ -2103,22 +2105,19 @@ pub async fn generate_request(request: &HttpRequest, mode: &GeneratorTestMode, c
           if let Some(parameter) = parameters.get_mut(param) {
             let mut generated = parameter.clone();
             for (index, val) in parameter.iter().enumerate() {
-              if let Ok(v) = generator.generate_value(val, context, &DefaultVariantMatcher.boxed()) {
-                generated[index] = v;
+              let value = val.clone().unwrap_or_default();
+              if let Ok(v) = generator.generate_value(&value, context, &DefaultVariantMatcher.boxed()) {
+                generated[index] = Some(v);
```
JP-Ellis, line 2110: As above, just double checking whether the loss of distinction between `Some("")` and `None` is intentional.

rust/pact_matching/src/query.rs
```diff
@@ -13,22 +13,24 @@ use crate::matchingrules::compare_lists_with_matchingrules;
...
 pub(crate) fn match_query_maps(
-  expected: HashMap<String, Vec<String>>,
-  actual: HashMap<String, Vec<String>>,
+  expected: HashMap<String, Vec<Option<String>>>,
+  actual: HashMap<String, Vec<Option<String>>>,
   context: &dyn MatchingContext
 ) -> HashMap<String, Vec<Mismatch>> {
   let mut result: HashMap<String, Vec<Mismatch>> = hashmap!{};
   for (key, value) in &expected {
+    let expected_value = value.iter().map(|v| v.clone().unwrap_or_default()).collect_vec();
```
JP-Ellis, line 22: And here again

rust/pact_matching/src/query.rs
```diff
@@ -13,22 +13,24 @@ use crate::matchingrules::compare_lists_with_matchingrules;
...
     match actual.get(key) {
       Some(actual_value) => {
-        let mismatches: Result<(), Vec<super::Mismatch>> = match_query_values(key, value, actual_value, context)
+        let actual_value = actual_value.iter().map(|v| v.clone().unwrap_or_default()).collect_vec();
```
JP-Ellis, line 25: Idem

rust/pact_matching/src/query.rs
```diff
@@ -40,7 +42,7 @@ pub(crate) fn match_query_maps(
       None => result.entry(key.clone()).or_default().push(Mismatch::QueryMismatch {
         parameter: key.clone(),
         expected: "".to_string(),
-        actual: format!("{:?}", value),
+        actual: format!("{:?}", value.iter().map(|v| v.clone().unwrap_or_default()).collect_vec()),
```
JP-Ellis, line 45: For debugging, it probably would be best to distinguish `Some("")` from `None`.

JP-Ellis (changes requested):
If I understand correctly, the internal representations are:

| Query String | Key | Value |
|-|-|-|
| `foo=bar` | `"foo"` | `Some("bar")` |
| `foo=` | `"foo"` | `Some("")` |
| `foo` | `"foo"` | `None` |

If so, I just noticed a few instances where `unwrap_or_default()` might be used a bit too eagerly, resulting in a loss of distinction between `Some("")` and `None` (see [this Rust Playground demo](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=98553e54ef9e8cf060b3e4e510456ec3))

I'm flagging them to make sure they are all intentional, and won't result in inconsistent behaviours across different parts of the codebase.

rust/pact_ffi/src/mock_server/handles.rs
rholshausen: I will copy your example docs :-D

compatibility-suite/tests/v3_steps/generators.rs
rholshausen: This is in the compatibility-suite, it doesn't matter

rust/pact_matching/src/lib.rs
rholshausen: Generator code, value doesn't matter.

rust/pact_ffi/src/mock_server/handles.rs
JP-Ellis: Nice, the only one to fill out would be the last one (as I didn't bother filling out the Integration JSON text) 👍

rust/pact_matching/src/query.rs
rholshausen: Just for display purposes, keeps previous behaviour

rust/pact_matching/src/query.rs
rholshausen: Just for display purposes, keeps previous behavior. The expected and actual values are not actually used by anything.

JP-Ellis (approved):

rust/pact_ffi/src/mock_server/handles.rs
rholshausen: Integration JSON won't support it, it will always resolve to the empty string.

PR #412, opened 2024-04-28
fix: build with glibc 2.28 for x86_64_gnu

praveen-em:
An attempt to fix https://github.com/pact-foundation/pact-js-core/issues/502 using cross.


YOU54F:
I think the failure before was due to cross re-using deps in the target folder

the macos failure, is indeed due to binary content matching, and is due to macos-latest switching to arm based runners.

we can revert back to original behaviour by switching the macos-latest tag to macos-12

praveen-em:
That makes sense! I saw your PR which is all green. I will close this one.

PR #429, opened 2024-05-28
feat: update content_type_matching logic for binary payloads

### LINKED ISSUE #171: Matching binary payloads does not work consistently across OS/Platform
Given a test that looks like this:

```js
      const interaction = pact.newInteraction('some description');

      interaction.uponReceiving('a request to create a dog with binary data');
      interaction.withRequest('POST', '/dogs/1234');
      interaction.withRequestBinaryBody(bytes, 'application/gzip');
      interaction.withStatus(200);

      port = pact.createMockServer(HOST);

      // and http request to that mock server:

      return axios
        .request({
          baseURL: `http://${HOST}:${port}`,
          headers: {
            'content-type': 'application/gzip',
          },
          data: bytes, 
          method: 'POST',
          url: '/dogs/1234',
        })
```

We get the following error:

> Expected header 'Content-Type' to have value 'application/octet-stream' but was 'application/gzip'

This is confusing and unexpected, because nowhere in my test setup was I expecting that type, and am sending the correct `content-type` header from the HTTP client.

If I change the test to instead use both `application/octet-stream` I get a different error:

> Expected binary contents to have content type 'application/octet-stream' but detected contents was 'application/gzip'. 

The assertions for the header are correct, but it's failing to match on the body. 

### Problem

We can make this work by modifying it so that the setup is this way:

```js
interaction.withRequestBinaryBody(bytes, 'application/gzip');
...
 return axios
        .request({
          baseURL: `http://${HOST}:${port}`,
          headers: {
            'content-type': 'application/octet-stream',
          },
          data: bytes,
          method: 'POST',
          url: '/dogs/1234',
        })
```

However, aside from that being a little confusing, it's not reliable on Windows, which expects the content type to be as per the initial setup.

The current consequence is that I can't write a small feature spec in Pact JS that works x-platform, but more broadly it seems like a bug.

#### 1. Disable `sys_fdo_magic` in tree_magic_mini crate

From: https://crates.io/crates/tree_magic_mini/3.0.3

> All mime information and relation information is loaded from the Shared MIME-info Database as described at https://specifications.freedesktop.org/shared-mime-info-spec/shared-mime-info-spec-latest.html. If you beleive that this is not present on your system, turn off the sys_fdo_magic feature flag.

If I understand that correctly, this would disable the reliance on behaviour of a third party library, and result in reliable behaviour in this crate.

#### 2. Use content types
In the presence of a `content-type` header, do not use the MIME type sniffing to determine the data type.

There was some other options proposed in this slack thread: https://pact-foundation.slack.com/archives/CA2S7E6KC/p1642474173003300

### Example builds

1. https://github.com/pact-foundation/pact-js/runs/4848121252?check_suite_focus=true#step:4:7668 (linux passing, windows failing)
2. https://github.com/pact-foundation/pact-js-core/runs/4812794228?check_suite_focus=true#step:4:361 (note the ubuntu build passes, but osx fails)

YOU54F:
fixes #171 

###### Content Type - Detection Mechanisms for Binary Content

`pact_matching` now performs the following for matching Binary content-types

1. Determines `expected` `Content-Type` header requested by user in test
2. Read content buffer with `infer` library, and guess `Content-Type` based on magic bytes
3. If unsuccessful
   1. Read content buffer with `tree_magic_mini` library, and guess `Content-Type` based on shared-mime-info DB
      1. MagicDB is not shipped with pact_matching, due to GPL restrictions, users can add manually
         1. Linux Alpine - `apk add shared-mime-info`
         2. MacOS `brew install shared-mime-info`
            1. `arm64` MacOS requires `tree_magic_mini` [fork](https://github.com/you54f/tree_magic) 
         3. Linux - Debian`apt-get install -y shared-mime-info`
4. If either result returns `text/plain`, then manually read bytes using `detect_content_type_from_bytes` function in `pact_models`
5. If all of `2`, `3`, or `4` fails, then throw error, otherwise return `Ok`

certainly better ways to implement this, but wanted to make sure I had a full round trip in all the consuming languages, and could remove the content-type hacks there.

The plus side is we have consistent behaviour in all of our test examples, on all supported platforms and architectures (win/lin/mac arm64 and amd64) without the use of tree_magic and the existence of shared-mime-info (which is great for container users, and great for avoiding GPL woes) 

Rust libraries used:

- https://github.com/mbrubeck/tree_magic
  - fork https://github.com/you54f/tree_magic (for m1 support), PR raised
- https://github.com/bojand/infer
- Explore (not used but worth a look at) - https://github.com/ebassi/xdg-mime-rs

Questions

1) Do we need tree_magic_mini, I don't seem to hit that check in the wild
2) We probably want some vendored format examples that are harder to detect
3) Provide user full opt-out of content-type matching, in case of things going weird?

Notes to merger

1) will need updating in pact_mock_server (use of pact_matching) which is then used in this repo, so I believe we will need to build/release pact_matching to crates.io, update pact_mock_server, release, and then back to pact_ref to do the works there

Tested in 

- pact rust
- pact rust compat suite
- pact-js-core
  - https://github.com/YOU54F/pact-js-core/pull/42
  - https://github.com/YOU54F/pact-js-core/pull/43
- pact-js
  - https://github.com/YOU54F/pact-js/pull/1
- pact-php
  - https://github.com/YOU54F/pact-php/pull/1
- pact-net
  - https://github.com/YOU54F/pact-net/pull/1 
- pact-ruby-ffi
  - https://github.com/YOU54F/pact-ruby-ffi/pull/3

### DIFF
--- a/compatibility-suite/pact-compatibility-suite/features/V3/matching_rules.feature
+++ b/compatibility-suite/pact-compatibility-suite/features/V3/matching_rules.feature
@@ -212,7 +212,7 @@ Feature: V3 era Matching Rules
       | application/octet-stream | file: sample.pdf |
     When the request is compared to the expected one
     Then the comparison should NOT be OK
-    And the mismatches will contain a mismatch with error "$" -> "Expected binary contents to have content type 'image/jpeg' but detected contents was 'application/pdf'"
+    And the mismatches will contain a mismatch with error "$" -> "Expected binary contents to have content type 'image/jpeg' but inferred contents are 'application/pdf', magic contents are 'application/pdf'"
 
   Scenario: Supports a Values matcher (positive case, ignores missing and additional keys)
     Given an expected request configured with the following:
--- a/compatibility-suite/pact-compatibility-suite/features/V3/matching_rules.feature
+++ b/compatibility-suite/pact-compatibility-suite/features/V3/matching_rules.feature
@@ -212,7 +212,7 @@ Feature: V3 era Matching Rules
       | application/octet-stream | file: sample.pdf |
     When the request is compared to the expected one
     Then the comparison should NOT be OK
-    And the mismatches will contain a mismatch with error "$" -> "Expected binary contents to have content type 'image/jpeg' but inferred contents are 'application/pdf', magic contents are 'application/pdf'"
+    And the mismatches will contain a mismatch with error "$" -> "Expected binary contents to have content type 'image/jpeg' but detected contents was 'application/pdf'"
 
   Scenario: Supports a Values matcher (positive case, ignores missing and additional keys)
     Given an expected request configured with the following:
--- a/rust/pact_consumer/Cargo.toml
+++ b/rust/pact_consumer/Cargo.toml
@@ -30,7 +30,7 @@ itertools = "0.12.1"
 lazy_static = "1.4.0"
 maplit = "1.0.2"
 pact_matching = { version = "~1.2.2", path = "../pact_matching", default-features = false }
-pact_mock_server = { version = "~1.2.6", default-features = false }
+pact_mock_server = { version = "=1.2.8", default-features = false, git = "https://github.com/you54f/pact-core-mock-server.git", branch = "main" }
 pact_models = { version = "~1.2.0", default-features = false }
 pact-plugin-driver = { version = "~0.6.1", optional = true, default-features = false }
 regex = "1.10.4"
--- a/rust/pact_ffi/Cargo.toml
+++ b/rust/pact_ffi/Cargo.toml
@@ -31,7 +31,8 @@ maplit = "1.0.2"
 multipart = { version = "0.18.0", default-features = false, features = ["client", "mock"] }
 onig = { version = "6.4.0", default-features = false }
 pact_matching = { version = "~1.2.2", path = "../pact_matching" }
-pact_mock_server = { version = "~1.2.6" }
+pact_mock_server = { version = "=1.2.8", git = "https://github.com/you54f/pact-core-mock-server.git", branch = "main" }
+# pact_mock_server = { version = "=1.2.8", path = "../../../pact-core-mock-server/pact_mock_server" }
 pact_models = { version = "~1.2.0" }
 pact-plugin-driver = { version = "~0.6.1" }
 pact_verifier = { version = "~1.2.1", path = "../pact_verifier" }
--- a/rust/pact_matching/README.md
+++ b/rust/pact_matching/README.md
@@ -510,6 +510,33 @@ Example:
 matching(contentType, 'application/json', '{}')
 ```
 
+###### Content Type - Detection Mechanisms for Binary Content
+
+`pact_matching` currently performs the following for matching Binary content-types
+
+1. Determines `expected` `Content-Type` header requested by user in test
+2. Read content buffer with `infer` library, and guess `Content-Type` based on magic bytes
+3. If unsuccessful
+   1. Read content buffer with `tree_magic_mini` library, and guess `Content-Type` based on shared-mime-info DB
+      1. MagicDB is not shipped with pact_matching, due to GPL restrictions, users can add manually
+         1. Linux Alpine - `apk add shared-mime-info`
+         2. MacOS `brew install shared-mime-info`
+            1. `arm64` MacOS requires `tree_magic_mini` [fork](https://github.com/you54f/tree_magic) 
+         3. Linux - Debian`apt-get install -y shared-mime-info`
+4. If either result returns `text/plain`, then manually read bytes using `detect_content_type_from_bytes` function in `pact_models`
+5. If all of `2`, `3`, or `4` fails, then throw error, otherwise return `Ok`
+
+Rust libraries used:
+
+- https://github.com/mbrubeck/tree_magic
+  - fork https://github.com/you54f/tree_magic
+- https://github.com/bojand/infer
+- Explore:- https://github.com/ebassi/xdg-mime-rs
+
+TODO:
+
+Provide user full opt-out of content-type matching?
+
 ##### Matching an example type by reference
 
 Type matching can also be specified by a reference to an example. References are defined by a dollar (`$`) followed by
--- a/rust/pact_matching/README.md
+++ b/rust/pact_matching/README.md
@@ -529,13 +529,7 @@ matching(contentType, 'application/json', '{}')
 Rust libraries used:
 
 - https://github.com/mbrubeck/tree_magic
-  - fork https://github.com/you54f/tree_magic
 - https://github.com/bojand/infer
-- Explore:- https://github.com/ebassi/xdg-mime-rs
-
-TODO:
-
-Provide user full opt-out of content-type matching?
 
 ##### Matching an example type by reference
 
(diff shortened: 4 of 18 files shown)


rholshausen:
Looks like infer will handle the most common types, but it has all the checks build in in Rust code, not config, so it is not extensible and will not be able to detect new types unless the crate is updated. So having tree_magic as a fallback is probably a good idea.

compatibility-suite/pact-compatibility-suite/features/V3/matching_rules.feature
```diff
@@ -212,7 +212,7 @@ Feature: V3 era Matching Rules
       | application/octet-stream | file: sample.pdf |
     When the request is compared to the expected one
     Then the comparison should NOT be OK
-    And the mismatches will contain a mismatch with error "$" -> "Expected binary contents to have content type 'image/jpeg' but detected contents was 'application/pdf'"
```
rholshausen, line 215: This change should be done in the compatibility suite repo, and it should only return one inferred type.

YOU54F: cheers for the heads up, ill revert the mismatch message back to its original state and add debug logging in the matching method to state which detection method we have tried

rust/pact_matching/src/json.rs
```diff
@@ -975,9 +975,8 @@ mod tests {
     let matcher = MatchingRule::ContentType("text/plain".to_string());
     expect!(Value::String("plain text".into()).matches_with(&Value::String("plain text".into()), &matcher, false)).to(be_ok());
     expect!(Value::String("plain text".into()).matches_with(&Value::String("different text".into()), &matcher, false)).to(be_ok());
-    expect!(Value::String("plain text".into()).matches_with(&json!(100), &matcher, false)).to(be_ok());
-    expect!(Value::String("plain text".into()).matches_with(&json!(100.01), &matcher, false)).to(be_ok());
-    #[cfg(not(windows))]
+    expect!(Value::String("plain text".into()).matches_with(&json!(100), &matcher, false)).to(be_err());
+    expect!(Value::String("plain text".into()).matches_with(&json!(100.01), &matcher, false)).to(be_err());
```
YOU54F, line 979: this is a change in behaviour. was this correct before? I think we should get an error, based on the condition of byte checking in this repo when content type is text/plain but the actual detected content differs.

rholshausen: This is the interesting thing about context types. JSON is a text format, so matching `text/plain` with a JSON document should be OK (same goes for other text formats). I.e. If I say I need to get plain text, and you give me JSON, HTML or a Gerhkin file, it is plain text.

rholshausen: Apache Tika (which Pact-JVM uses) solves this by creating a hierarchy for the content types. So the parent of `application/json` is `text/plain`.

YOU54F: My logic was incorrect, and not the same as the original impl for this case, so it was a bug well caught

YOU54F: >Apache Tika (which Pact-JVM uses) solves this by creating a hierarchy for the content types. So the parent of application/json is text/plain.

I laugh now, but I wonder at which point we end up needing to port the logic to Rust, even if just to get consistent behaviour between our two cores

rholshausen: I have thought about that many times, but it is not a trivial endeavor.

YOU54F:
👋🏾 Tagged a few of you for review!

Always room for improvement on what is proposed but this gives us 

- consistent cross platform and cross arch content-type detection with the new crate for the existing raised issues
- falls back to the existing behaviour we have today, with additional improvements to also add support for arm64 macos in the tree_magic_mini crate for `shared-mime-info` db lookups
- improved debug logging, so we are aware of which content type lookups are being performed.

The compatibility suite has been updated to test cross platform (windows x86_64 + macos x86_64/arm64 ) and skipped tests for windows have been been re-instated 👍🏾 

Happy to make any amendments, as I know this might not be the best solution, especially with respect to my coding style, or the overall (existing) approach to matching in this function, but please be reminded this is a massive improvement over our existing confusing behaviour, and we can always progress forwards.

rust/pact_matching/src/json.rs
YOU54F: hmm lets cross that bridge if and when we need to come to it!

rust/pact_consumer/Cargo.toml
```diff
@@ -30,7 +30,7 @@ itertools = "0.12.1"
 lazy_static = "1.4.0"
 maplit = "1.0.2"
 pact_matching = { version = "~1.2.2", path = "../pact_matching", default-features = false }
-pact_mock_server = { version = "~1.2.6", default-features = false }
+pact_mock_server = { version = "=1.2.8", default-features = false, git = "https://github.com/you54f/pact-core-mock-server.git", branch = "main" }
```
YOU54F, line 33: please update to released pact_mock_server, when pact_matching is released, and pact_mock_server updated to reference released crate

rust/pact_ffi/Cargo.toml
```diff
@@ -31,7 +31,8 @@ maplit = "1.0.2"
 multipart = { version = "0.18.0", default-features = false, features = ["client", "mock"] }
 onig = { version = "6.4.0", default-features = false }
 pact_matching = { version = "~1.2.2", path = "../pact_matching" }
-pact_mock_server = { version = "~1.2.6" }
+pact_mock_server = { version = "=1.2.8", git = "https://github.com/you54f/pact-core-mock-server.git", branch = "main" }
```
YOU54F, line 34: please update to released pact_mock_server, when pact_matching is released, and pact_mock_server updated to reference released crate

rust/pact_ffi/Cargo.toml
```diff
@@ -31,7 +31,8 @@ maplit = "1.0.2"
...
+# pact_mock_server = { version = "=1.2.8", path = "../../../pact-core-mock-server/pact_mock_server" }
```
YOU54F, line 35: path useful for dev guidelines if users are working across both repos locally.

```
.
├── pact-core-mock-server
└── pact-reference

```

rholshausen:
Just a note, if you want to override a dependency, the easiest way is to use the `[patch.crates-io]` section in the workspace Cargo.toml

tienvx (commented):
LGTM.

I wonder: since this PR is merged, when can we revert custom changes related to the `pact_mock_server`?

rholshausen:
Two tests are failing on Windows: https://github.com/pact-foundation/pact-reference/actions/runs/9295821959/job/25583357274

rholshausen:
> Two tests are failing on Windows: https://github.com/pact-foundation/pact-reference/actions/runs/9295821959/job/25583357274

All good. Mock server is using the old pact matching. Updated it and ran the tests on Windows, all pass.

rust/pact_matching/README.md
```diff
@@ -510,6 +510,27 @@ Example:
...
+`pact_matching` currently performs the following for matching Binary content-types
+
+1. Determines `expected` `Content-Type` header requested by user in test
+2. Read content buffer with `infer` library, and guess `Content-Type` based on magic bytes
+3. If unsuccessful
+   1. Read content buffer with `tree_magic_mini` library, and guess `Content-Type` based on shared-mime-info DB
+      1. MagicDB is not shipped with pact_matching, due to GPL restrictions, users can add manually
+         1. Linux Alpine - `apk add shared-mime-info`
+         2. MacOS `brew install shared-mime-info`
+            1. `arm64` MacOS requires `tree_magic_mini` [fork](https://github.com/you54f/tree_magic)
```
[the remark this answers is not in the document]
YOU54F: it has been yes! i need to update that comment to reflect, we are consuming 3.1.5 in here now

PR #432, opened 2024-06-05
chore(ci): refactor indiv jobs

### LINKED ISSUE #436: 0.4.20 pact_ffi - aarch64-musl .so failing to load (missing symbols)
So we have an issue with the pact_ffi shared library, for aarch64-musl `.so` in `0.4.20`

- [0.4.19 passing](https://github.com/YOU54F/pact-ruby-ffi/actions/runs/9087464050) when integrated with ruby (loads the `.so` file)
- [0.4.20 failing](https://github.com/YOU54F/pact-ruby-ffi/actions/runs/9087466381)

Error

```
LoadError:
  Could not open library '/app/lib/pact/../../ffi/linux-arm64-musl/libpact_ffi.so': Error relocating /app/lib/pact/../../ffi/linux-arm64-musl/libpact_ffi.so: __builtin_copysignq: symbol not found.
  Searched in <system library path>, /usr/lib, /usr/local/lib, /opt/local/lib
# /usr/local/bundle/gems/ffi-1.16.3/lib/ffi/dynamic_library.rb:65:in `load_library'
```

Seems to be exclusive to aarch64, and the dynamic library only. amd64 tests are passing in pact-php as well using the shared library, and the static .a is being used and tested successfully in pact-js-core in this [PR](https://github.com/pact-foundation/pact-js-core/pull/507)

First reported by @tienvx over in pact-php alpine [PR](https://github.com/pact-foundation/pact-php/pull/613#discussion_r1599747130) - Thanks Tien!

YOU54F:
Building individual targets allows us to identify where a single one fails, and gives us quicker feedback.

This mitigates some of the issues seen when building multiple targets with cross (wrong linker symbols), and allows us to enable per target caching which improves compile times.

Few other quality of life improvements are included (cron job, smoke test to weed out loading issues with the shared libraries)

- update release workflow to support
  - workflow_dispatch
  - cron schedule every wednesday
  - run each target in seperate job
  - add cache key per target

- update release scripts (pact_ffi / verifier_cli )
  - allow building individual targets
  - refactor similar steps
  - generate musl .so with `RUSTFLAGS="-C target-feature=-crt-static"` fixes #436 
  - smoke test dynamic lib loading (for pact-php, pact-js, pact-net, pact-go) fixes #436 
  
## Breaking changes
  
BREAKING CHANGE: macos aarch64 binaries now renamed to <app_name>-macos-aarch64. (previously was apple-darwin for aarch64)

Rationale: for consistency with x86_64 apple target and other platforms

BREAKING CHANGE: macos binaries are now only published with <app_name>-macos-<arch>. `osx` flavoured varieties are now discontinued

Rationale: for consistency with nomenclature, as osx not relevant today with latest version being macos 14

BREAKING CHANGE: sha256 files no longer contain full paths to files, only filenames

Rationale: The additional information is superfluous for end users, and requires them to modify the file for it to be usable. [see pact-net](https://github.com/pact-foundation/pact-net/blob/09ee8189a3dba1c47a02e571e8872924b90caaa8/build/download-native-libs.sh#L43-L50)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,22 +9,38 @@ on:
     branches:
       - master
 
+  workflow_dispatch:
+
+  schedule:
+    - cron: 0 0 * * 4 # Midnight Wednesday
+
 concurrency:
   group: release-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
   build-release:
+    name: build-release (${{ matrix.targets }})
     runs-on: ${{ matrix.operating-system }}
     strategy:
       matrix:
         include:
           - operating-system: ubuntu-20.04
-            targets: x86_64-unknown-linux-gnu,x86_64-unknown-linux-musl,aarch64-unknown-linux-gnu,aarch64-unknown-linux-musl
+            targets: x86_64-unknown-linux-gnu
+          - operating-system: ubuntu-20.04
+            targets: x86_64-unknown-linux-musl
+          - operating-system: ubuntu-20.04
+            targets: aarch64-unknown-linux-gnu
+          - operating-system: ubuntu-20.04
+            targets: aarch64-unknown-linux-musl
           - operating-system: windows-2019
-            targets: aarch64-pc-windows-msvc,x86_64-pc-windows-msvc
+            targets: x86_64-pc-windows-msvc
+          - operating-system: windows-2019
+            targets: aarch64-pc-windows-msvc
           - operating-system: macos-12
-            targets: aarch64-apple-darwin,x86_64-apple-darwin
+            targets: x86_64-apple-darwin
+          - operating-system: macos-14
+            targets: aarch64-apple-darwin
       fail-fast: false
 
     env:
@@ -43,6 +59,7 @@ jobs:
         uses: Swatinem/rust-cache@v2
         with:
           workspaces: rust
+          key: ${{ matrix.targets }}
 
       - name: Set up QEMU
         if: runner.os == 'Linux'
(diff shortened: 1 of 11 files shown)


rholshausen:
I hope you know what you are doing!

.github/workflows/release.yml
```diff
@@ -71,52 +88,58 @@ jobs:
         id: cargo-flags
         shell: bash
         run: |
-          if [[ "${{ github.event_name }}" = "release" ]]; then
-            echo "flags=--release" >> "$GITHUB_OUTPUT"
-          else
-            echo "flags=" >> "$GITHUB_OUTPUT"
-          fi
-
+          echo "flags=--release" >> "$GITHUB_OUTPUT"
```
JP-Ellis, line 91: Since release builds can be significantly slower, should this still be only selectively enabled?

YOU54F: so originally when I was able to build for multi targets with cross in a single run, we would actually run out of disk space because the debug build was so large.

this workflow is only for building the released artifacts, until this PR where it does the most basic of smoke tests.

the actual build ffi workflow builds and tests the ffi lib cross plat (although not all of our targets, just x86_64 win/lin/mac)

so the rationale was if this is able to smoke test a full release wf before each target is actually released (so via the PR workflow), we actually know whether it will work when we push a release tag.

Maybe we could just run debug on the pr's, but run in release mode on master, so before someone does push a release tag, they have confidence it will just work (tm) 

I feel like we could actually probably consolidate some of our workflows to avoid de-depulication, and maximise our caching potential. we have a pact rust workflow which tests everything, pact ffi workflow that builds the project with cmake (which we don't use in the release workflow), and c smoke tests on linux against.

YOU54F:
> I hope you know what you are doing!

😅 

I've given it a good smoke test across all the projects consuming it that are in my immediate radar (pact-js-core / pact-php / pact-net / pact-ruby-ffi / pact-go )

It's certainly reduced my feedback loop and helped me diagnose single target failures, and hopefully this should mitigate some of the initial concerns we get from maintainers on a new update. it's bork and we don't know, and have to go through the release shenanigans again with a fix that may or may not work (when its something in the cross compilation build process, that is a bit of a black box without being prepared to go really deep in yak shave territory)

.github/workflows/release.yml
JP-Ellis: If we cache the `CARGO_TARGET_DIR`, we would likely get most of the speed up anyway. The compiler would not need to recompile most of the dependencies in the first place.

PR #435, opened 2024-06-08
Issue/434 missing buffer logs

### LINKED ISSUE #434: 0.4.20 pactffi_mock_server_logs empty - mismatch in tracing library
pactffi_mock_server_logs are empty, in 0.4.20 release.

Initially identified in pact-net's test suite when attempting to update to libpact_ffi 0.4.20. pact-net uses pactffi_log_to_buffer when it initialises, and has a test to check that pactffi_mock_server_logs actually returns logs (and it failed)

## Cause

I believe this is due to a mis-match in the logging libraries in `pact_models`, compared to the rest of the pact rust estate

- `tracing-0.1.37` https://github.com/pact-foundation/pact-reference/blob/master/rust/pact_models/Cargo.toml#L48
- `tracing-subscriber-0.3.16` https://github.com/pact-foundation/pact-reference/blob/5699cab757e31875b74ea8ce98cd9821dc331421/rust/pact_models/Cargo.toml#L67C1-L67C93

All other crates

- https://github.com/pact-foundation/pact-reference/blob/master/rust/pact_ffi/Cargo.toml#L49-L51
- https://github.com/pact-foundation/pact-reference/blob/5699cab757e31875b74ea8ce98cd9821dc331421/rust/pact_consumer/Cargo.toml#L39-L40
- https://github.com/pact-foundation/pact-reference/blob/5699cab757e31875b74ea8ce98cd9821dc331421/rust/pact_verifier/Cargo.toml#L48-L49
- https://github.com/pact-foundation/pact-reference/blob/5699cab757e31875b74ea8ce98cd9821dc331421/rust/pact_verifier_cli/Cargo.toml#L40-L43
- https://github.com/pact-foundation/pact-core-mock-server/blob/196cabbca4767df999e15f0aacb508549ba377a7/pact_mock_server/Cargo.toml#L43-L44
- https://github.com/pact-foundation/pact-plugins/blob/5a8e1df161fe7707146ea06b31588b35e9af2d08/drivers/rust/driver/Cargo.toml#L52-L53

## Reproductions

Added a test, which requires a buffer log sink. I believe the full test run sets a different global sink, so the test is ignored in the main run, and executed in isolation. It's not terribly pretty but it does the job.

✅ tag `libpact_ffi-v0.4.19`
 - Delta https://github.com/pact-foundation/pact-reference/compare/libpact_ffi-v0.4.19...YOU54F:pact-reference:issue/missing_logs_0.4.19
 - Passing Run https://github.com/YOU54F/pact-reference/actions/runs/9422764373/job/25959743140#step:6:121

❌ tag `libpact_ffi-v0.4.20`
 - Delta https://github.com/pact-foundation/pact-reference/compare/libpact_ffi-v0.4.20...YOU54F:pact-reference:issue/missing_logs_0.4.20
 - Failing Run https://github.com/YOU54F/pact-reference/actions/runs/9422936793/job/25960271309#step:6:66

## Resolution

Suggest

1. Update pact_models to match tracing libs
  1. Requires consuming in pact-plugin-driver & pact-core-mock-server
1. Pin logging libs explicitly with `version="=x.y.z"`, note the `=` sign in the version number. This forces pinned resolution

## Testing

1. Update pact_models to same logging libs
2. Update pact_mock_server to use updated pact_models
3. Update pact-plugin-driver, as it consumes pact_models, and are depended on by pact_mock_server
4. Update all the things in pact-ref to point to updated pact_mock_server and pact-plugin-driver

### pact-plugin-driver

- Delta https://github.com/pact-foundation/pact-plugins/compare/main...YOU54F:pact-plugins:fix/logging_min?expand=1
- Passing Run https://github.com/YOU54F/pact-plugins/actions/runs/9423940849/job/25963273163 (failures due to no token to publish to PactFlow)

### pact_mock_server

- Delta https://github.com/pact-foundation/pact-core-mock-server/compare/main...YOU54F:pact-core-mock-server:fix/logging_min?expand=1
- Passing Run https://github.com/YOU54F/pact-core-mock-server/actions/runs/9423753133 

### pact_reference

- Delta https://github.com/pact-foundation/pact-reference/compare/master...YOU54F:pact-reference:fix_logging?expand=1
- Passing run https://github.com/YOU54F/pact-reference/actions/runs/9423808069/job/25962909225

YOU54F:
fixes #434 

I believe this issue stems from one of the pact crates having a different version of the tracing crate used for logging.

This seems like an awkward position to be in, when upgrading dependencies of each of the indiv pact rust crates, especially as two sit across different repos (pact_mock_server / pact-plugin-driver).

Would it make sense to crate a pact_logging crate that wraps the required deps for logging, and this singular crate can be imported by all the other pact rust crates.

If we need to update the logging deps, we update them in the pact_logging crate, and then can use the workspace override to patch all crates to use the same versions (via pact_logging)

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -12,3 +12,7 @@ resolver = "2"
 
 [patch.crates-io]
 onig = { git = "https://github.com/rust-onig/rust-onig", default-features = false }
+pact_models = { path = "./pact_models" }
+pact_matching = { path = "./pact_matching" }
+pact_mock_server = { git = 'https://github.com/pact-reference/pact-core-mock-server.git', branch = "issue/134_missing_buffer_logs" }
+pact-plugin-driver = { git = 'https://github.com/pact-reference/pact-plugins.git', branch = "issue/134_missing_buffer_logs" }
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -14,5 +14,5 @@ resolver = "2"
 onig = { git = "https://github.com/rust-onig/rust-onig", default-features = false }
 pact_models = { path = "./pact_models" }
 pact_matching = { path = "./pact_matching" }
-pact_mock_server = { git = 'https://github.com/pact-reference/pact-core-mock-server.git', branch = "issue/134_missing_buffer_logs" }
-pact-plugin-driver = { git = 'https://github.com/pact-reference/pact-plugins.git', branch = "issue/134_missing_buffer_logs" }
+pact_mock_server = { git = 'https://github.com/pact-foundation/pact-core-mock-server.git', branch = "issue/134_missing_buffer_logs" }
+pact-plugin-driver = { git = 'https://github.com/pact-foundation/pact-plugins.git', branch = "issue/134_missing_buffer_logs" }
--- a/rust/pact_ffi/tests/tests.rs
+++ b/rust/pact_ffi/tests/tests.rs
@@ -9,7 +9,9 @@ use bytes::Bytes;
 use expectest::prelude::*;
 use itertools::Itertools;
 use libc::c_char;
+use log::LevelFilter;
 use maplit::*;
+use pact_ffi::log::pactffi_log_to_buffer;
 use pact_models::bodies::OptionalBody;
 use pact_models::PactSpecification;
 use pretty_assertions::assert_eq;
@@ -19,14 +21,14 @@ use tempfile::TempDir;
 use serde_json::{json, Value};
 use rstest::rstest;
 use regex::Regex;
-
 #[allow(deprecated)]
 use pact_ffi::mock_server::{
   pactffi_cleanup_mock_server,
   pactffi_create_mock_server,
   pactffi_create_mock_server_for_pact,
   pactffi_mock_server_mismatches,
-  pactffi_write_pact_file
+  pactffi_write_pact_file,
+  pactffi_mock_server_logs,
 };
 #[allow(deprecated)]
 use pact_ffi::mock_server::handles::{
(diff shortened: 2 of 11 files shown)


rust/Cargo.toml
```diff
@@ -12,3 +12,7 @@ resolver = "2"
 
 [patch.crates-io]
 onig = { git = "https://github.com/rust-onig/rust-onig", default-features = false }
+pact_models = { path = "./pact_models" }
+pact_matching = { path = "./pact_matching" }
+pact_mock_server = { git = 'https://github.com/pact-foundation/pact-core-mock-server.git', branch = "issue/134_missing_buffer_logs" }
```
YOU54F, line 17: note these are reliant on patches in branches which pull in the updated pact-models code from this pr.

will need co-ordinating with release of pact models, then pact-plugin-driver, which is required by pact-core-mock-server, which is required by crates in this repo.

rust/pact_ffi/tests/tests.rs
```diff
@@ -1406,3 +1408,30 @@ fn matching_definition_expressions_matcher() {
...
+    .body(r#"{"foo":"no-very-bar"}"#)
+    .send().expect("Sent POST request to mock server");
+
+  let logs =  unsafe {
+    CStr::from_ptr(pactffi_mock_server_logs(port)).to_string_lossy().into_owned()
+  };
+  println!("{}",logs);
+  assert_ne!(logs,"", "logs are empty");
+
+  pactffi_cleanup_mock_server(port);
```
rholshausen, line 1436: If the assertion above it fails, this line will not be executed.

rholshausen:
I'm wondering if we use the patch block to force the logging crate version, this will fix this issue. It does mean we need some way of working out when to update it. But the logging crates have been quite stable.

rholshausen:
I'm going to close this PR, as it can't be merged in the current state. I'll then cherry pick the changes from your branch.

rholshausen:
This was not caused by the logging crates, but by the FFI InMemory sink which uses the global static `LOG_BUFFER` in the pact_matching crate. This is required to be the same version across all crates (i.e. mock server crate).

YOU54F:
ahhh great, good detective work @rholshausen

PR #437, opened 2024-06-08
fix: merge pact duplicates with diff desc/same prov state

### LINKED ISSUE #389: Pact Merging Creates Duplicate Interactions
**FFI Version**: 0.4.16
**Specification Version**: 4

I have a test in PactNet which is very simple, like this:

```csharp
this.pact
    .UponReceiving("a request for an order with an unknown ID")
        .WithRequest(HttpMethod.Get, "/api/orders/404")
        .WithHeader("Accept", "application/json")
    .WillRespond()
        .WithStatus(HttpStatusCode.NotFound);
```

When I run the test, the mock server ends up being invoked with `overwrite:false` via the `pactffi_pact_handle_write_file` method.

**Expected**: The first time the test is executed then the new file will be written then every subsequent execution should just end up producing the same interaction definition (and thus merging it into the file effectively as a no-op).

**Actual**: Every time I run the test the same interaction is written as an exact duplicate, so my Pact file contains exactly the same thing three times in a row:

```json
{
  "description": "a request for an order with an unknown ID",
  "pending": false,
  "request": {
    "headers": {
      "Accept": [
        "application/json"
      ]
    },
    "method": "GET",
    "path": "/api/orders/404"
  },
  "response": {
    "status": 404
  },
  "type": "Synchronous/HTTP"
},
{
  "description": "a request for an order with an unknown ID",
  "pending": false,
  "request": {
    "headers": {
      "Accept": [
        "application/json"
      ]
    },
    "method": "GET",
    "path": "/api/orders/404"
  },
  "response": {
    "status": 404
  },
  "type": "Synchronous/HTTP"
},
{
  "description": "a request for an order with an unknown ID",
  "pending": false,
  "request": {
    "headers": {
      "Accept": [
        "application/json"
      ]
    },
    "method": "GET",
    "path": "/api/orders/404"
  },
  "response": {
    "status": 404
  },
```

YOU54F:
fix: merge message / sync pacts with no provider state

fixes #389


rust/pact_ffi/tests/tests.rs
```diff
@@ -1182,6 +1183,405 @@ fn repeated_interaction() {
...
+  }
+}"#
+  );
+}
+
+// Issue - Should be able to set version of message pact, and write to file containing v4 interactions
+// seems to be a problem setting the version, which defaults to v3
+// pactffi_new_message will accept a MessageHandle over the FFI barrier, but rust typing wont allow us
+// to pass a PactHandle along with pactffi_with_specification
+#[ignore = "require ability to set pact specification version in pactffi_new_message_pact"]
```
YOU54F, line 1459: i probably need to pull this out into a new issue. pact-new uses a new_message_interaction which is replicated in the test above.

YOU54F:
See pact-stub-server for an example

https://github.com/pact-foundation/pact-stub-server/blob/535d05cab267702da9dbc33fe8f856b02bbd610a/.github/workflows/release.yml#L77

PR #438, opened 2024-06-09
feat(verifier): support fallback_branch consumer version selector

### LINKED ISSUE #195: Consumer Version Selector `fallbackBranch` Is Ignored
The `fallbackBranch` option on consumer version selectors is ignored, even though it's a [valid property](https://docs.pact.io/pact_broker/advanced_topics/consumer_version_selectors#properties). 

If you run:

```bash
pact_verifier_cli \
    -b "https://pactbroker.example.org" \
    --consumer-version-selectors '{"branch":"feat/foo","fallbackBranch":"main"}' \
    -n "My API" \
    -l debug
```

then the verifier submits the following:

```
[DEBUG] (1) pact_verifier::pact_broker: Sending JSON to https://pactbroker.example.org/pacts/provider/My%20API/for-verification
using POST: {"providerVersionTags":[],"includePendingStatus":false,"consumerVersionSelectors":[{"branch":"feat/foo"}]}
```

This then retrieves no pacts if the consumers aren't using the same branch name, and so the provider test fails:

```
Failures:

1) Failed to load pact - Could not load pacts from the pact broker 'https://pactbroker.example.org'
- NotFound("No pacts were found for this provider")
```

YOU54F:
fixes #195 - also documents consumer version selector options as per pact broker docs

### DIFF
--- a/rust/pact_verifier/src/pact_broker.rs
+++ b/rust/pact_verifier/src/pact_broker.rs
@@ -1005,27 +1005,61 @@ async fn publish_provider_branch(
 /// Structure to represent a HAL link
 pub struct ConsumerVersionSelector {
   /// Application name to filter the results on
+  /// allows a selector to only be applied to a certain consumer.
   pub consumer: Option<String>,
   /// Tag
+  /// the tag name(s) of the consumer versions to get the pacts for. 
+  /// *This field is still supported but it is recommended to use the `branch` in preference now.*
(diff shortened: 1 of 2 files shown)


rust/pact_verifier/src/pact_broker.rs
rholshausen, line 1: Nice! Lots of comments.

PR #443, opened 2024-06-19
feat: add `generate_contents` methods

JP-Ellis:
The various `get_contents` FFI method return the raw (unprocessed) contents of the request/response, including all generators and matching rules. The `generate_contents` alternatives are used to process these generators so that the actual messages as received by the consumer can be retrieved.

### DIFF
--- a/rust/pact_ffi/src/models/async_message.rs
+++ b/rust/pact_ffi/src/models/async_message.rs
@@ -1,10 +1,15 @@
 //! V4 ASynchronous messages
 
+use std::collections::HashMap;
+
 use anyhow::anyhow;
 use bytes::Bytes;
+use futures::executor::block_on;
 use libc::{c_char, c_int, c_uchar, c_uint, EXIT_FAILURE, EXIT_SUCCESS, size_t};
+use pact_matching::generators::apply_generators_to_async_message;
 use pact_models::bodies::OptionalBody;
 use pact_models::content_types::{ContentType, ContentTypeHint};
+use pact_models::generators::GeneratorTestMode;
 use pact_models::provider_states::ProviderState;
 use pact_models::v4::async_message::AsynchronousMessage;
 use pact_models::v4::message_parts::MessageContents;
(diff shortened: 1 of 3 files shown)


rust/pact_ffi/src/models/async_message.rs
```diff
@@ -59,6 +64,41 @@ ffi_fn! {
...
+    ///
+    /// # Safety
+    ///
+    /// The data pointed to by the pointer must be deleted with
+    /// [`pactffi_message_contents_delete`][crate::models::contents::pactffi_message_contents_delete]
+    ///
+    /// # Error Handling
+    ///
+    /// If the message is NULL, returns NULL.
+    fn pactffi_async_message_generate_contents(message: *const AsynchronousMessage) -> *mut MessageContents {
```
rholshausen, line 84: Why return a mutable pointer? What will you want to change on the contents after the call?

rust/pact_ffi/src/models/sync_message.rs
```diff
@@ -437,6 +477,46 @@ ffi_fn! {
...
+        let plugin_data = Vec::new();
+        let interaction_data = HashMap::new();
+        let (_, mut responses) = block_on(apply_generators_to_sync_message(
+            &message,
+            &GeneratorTestMode::Consumer,
+            &context,
+            &plugin_data,
+            &interaction_data,
+        ));
+        if index >= responses.len() {
```
rholshausen, line 510: Optimisation: Check the length before doing the work to generate the responses

rust/pact_ffi/src/models/async_message.rs
JP-Ellis: The reason for the mutable pointer is that the underlying `MessageContents` is not owned by anything else, with the pointer being the only reference to it. The eventual deletion of the message contents would 'mutate' the message (in so far as deletion is a 'mutation').

I also don't see any risk in allowing people to mutate the `MessageContents`, though I also don't see a use case for that.

JP-Ellis: Also, the `ptr::raw_to(contents)` function returns a mutable pointer, so I would either need to cast it to a non-mutable pointer; or use another function.

rholshausen: In other places it casts to a const pointer, and then casts back to call the drop function. Just makes the intent clear, but is not a biggy. For instance https://github.com/pact-foundation/pact-reference/blob/1dce67a50148a2671b515dfb07aae076e08777aa/rust/pact_ffi/src/models/consumer.rs#L61

JP-Ellis:
Fixed the points you mentioned above.

I'll add a few tests as well to make sure the functions are working as expected.

JP-Ellis:
Tests have now been added.

PR #444, opened 2024-06-19
feat: add with_generators

JP-Ellis:
Add FFI to add generators without requiring the use of integration JSON.

### DIFF
--- a/rust/pact_ffi/src/mock_server/handles.rs
+++ b/rust/pact_ffi/src/mock_server/handles.rs
@@ -120,7 +120,7 @@ use maplit::*;
 use pact_models::{Consumer, PactSpecification, Provider};
 use pact_models::bodies::OptionalBody;
 use pact_models::content_types::{ContentType, detect_content_type_from_string, JSON, TEXT, XML};
-use pact_models::generators::{Generator, Generators};
+use pact_models::generators::{generators_from_json, Generator, Generators};
 use pact_models::headers::parse_header;
 use pact_models::http_parts::HttpPart;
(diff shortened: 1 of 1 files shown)


JP-Ellis:
Before merging, I just want to add a few tests first; though I can confirm that it works locally in Pact Python.

rust/pact_ffi/src/mock_server/handles.rs
```diff
@@ -2119,6 +2119,93 @@ ffi_fn!{
   }
 }
 
+ffi_fn!{
+  /// Add generators to the interaction.
+  ///
+  /// * `interaction` - Interaction handle to set the generators for.
+  /// * `part` - Request or response part (if applicable).
+  /// * `rules` - JSON string of the generators to add to the interaction.
```
rholshausen, line 2127: Comment doesn't match the parameter names

rust/pact_ffi/src/mock_server/handles.rs
```diff
@@ -2119,6 +2119,93 @@ ffi_fn!{
...
+  /// # Safety
+  ///
+  /// The rules parameter must be a valid pointer to a NULL terminated UTF-8
+  /// string.
+  fn pactffi_with_generators(
+    interaction: InteractionHandle,
+    part: InteractionPart,
+    generators: *const c_char
+  ) -> bool {
+    let generators = match convert_cstr("rules", generators) {
```
rholshausen, line 2145: should be generators not rules

JP-Ellis:
Fixed the typos, and added a test. Ready for review :)

PR #456, opened 2024-07-16
fix(ffi): shared processing of body

JP-Ellis:
The logic of processing the body within the `pactffi_with_body` was not consistent across the various interaction types. This refactors the logic out into a separate `process_body` function.

Note that as a result of this change, the behaviour of the FFI _may_ change.

### DIFF
--- a/rust/pact_ffi/src/mock_server/handles.rs
+++ b/rust/pact_ffi/src/mock_server/handles.rs
@@ -119,7 +119,7 @@ use libc::{c_char, c_int, c_uint, c_ushort, EXIT_FAILURE, EXIT_SUCCESS, size_t};
 use maplit::*;
 use pact_models::{Consumer, PactSpecification, Provider};
 use pact_models::bodies::OptionalBody;
-use pact_models::content_types::{ContentType, detect_content_type_from_string, JSON, TEXT, XML};
+use pact_models::content_types::{detect_content_type_from_string, ContentType, TEXT};
 use pact_models::generators::{generators_from_json, Generator, Generators};
 use pact_models::headers::parse_header;
 use pact_models::http_parts::HttpPart;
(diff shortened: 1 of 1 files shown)


rust/pact_ffi/src/mock_server/handles.rs
```diff
@@ -1650,185 +1650,224 @@ pub extern fn pactffi_response_status_v2(interaction: InteractionHandle, status:
...
+        matching_rules,
+        generators
+    );
+    let detected_type = detect_content_type_from_string(body);
+    let content_type = content_type
+        .clone()
+        .or_else(|| content_type_hint.clone())
+        .or_else(|| detected_type.clone())
+        .or_else(|| Some(TEXT.clone()));
+    trace!(?detected_type);
```
rholshausen, line 1688: Better to combine these two into a single trace

rust/pact_ffi/src/mock_server/handles.rs
```diff
@@ -1650,185 +1650,224 @@ pub extern fn pactffi_response_status_v2(interaction: InteractionHandle, status:
...
+                        None,
+                    )
+                }
+                _ => {
+                    trace!("Raw XML body left as is");
+                    OptionalBody::from(body)
+                }
+            }
+        }
+        None | Some(_) => {
```
rholshausen, line 1727: This can just be abbreviated as `_ => {`

rholshausen (approved):

rust/pact_ffi/src/mock_server/handles.rs
JP-Ellis: Accidental left-over of some refactoring. Fixed :)

JP-Ellis:
Fixed the issue with the failing tests. Reason being that the `Content-Type` header was not being set if `header` were `None`. The new behaviour is:

- If `headers` is `None`, create a `HashMap` and populate it
- If the `headers` is `Some(hm)`, then insert the new `Content-Type` header only if it does not exist already. This means that if someone has explicitly set the content type header, this function will _not_ override it.

JP-Ellis:
Docstring has been updated. @rholshausen, I'll leave you do a last check before merging 👍

rholshausen (approved):

YOU54F:
I believe this change has possibly caused a regression in the pact-php compat and test suite, see mentioned [PR](https://github.com/pact-foundation/pact-php/pull/619).

Changes also consumed in following PR's.

- [pact-go](https://github.com/pact-foundation/pact-go/pull/446)
- [pact-js-core](https://github.com/pact-foundation/pact-js-core/pull/519)

There are xml tests in pact-js but they are only consumable on release of pact-js-core (unless locally tested)

JP-Ellis:
Hmm, I was worried this change may have unintended repercussions, though was relieved to see the test suite to pass... Evidently an edge case was not covered.

I'll have a look at the PR that uncovered this unintended change, though given neither Pact Go or Pact JS picked up this issue, I'm wondering whether Tien may have unintentionally baked in the broken behaviour into a test?

Once I've figured out exactly the extent of the change, I'll update the changelog to highlight this change in behaviour in case it impacts anyone else in the future.

YOU54F:
Ahh so I don't believe pact-js-core or pact-go covered it. 

I think the detect content type from string is saying its application/json and it isn't being transformed into xml

I think its this bit of code

https://github.com/pact-foundation/pact-reference/pull/456/files#diff-d2ed99251aebcb3c325d1e2650a83cec3b1a8d470d9b524784b0797073b47fbdL1751-L1758

And I believe that we need to check if its actually xml, when intermediate json is detected

https://github.com/pact-foundation/pact-reference/pull/456/files#diff-d2ed99251aebcb3c325d1e2650a83cec3b1a8d470d9b524784b0797073b47fbdR1696

JP-Ellis:
The logic should be fine.

There's two content types set at the start of that function:

1. `detected_type` which is taken from the body string, and
2. `content_type` which is taken from the function argument, with a series of fallbacks.

https://github.com/pact-foundation/pact-reference/blob/111a18cd4ef93b0a9337f5a3c31d22b5949fa6a3/rust/pact_ffi/src/mock_server/handles.rs#L1682-L1687

The main matching is done on the `content_type`:

https://github.com/pact-foundation/pact-reference/blob/111a18cd4ef93b0a9337f5a3c31d22b5949fa6a3/rust/pact_ffi/src/mock_server/handles.rs#L1694

Then if _that_ is XML, it checks if the body is JSON, and fallbacks to raw body parsing:

https://github.com/pact-foundation/pact-reference/blob/111a18cd4ef93b0a9337f5a3c31d22b5949fa6a3/rust/pact_ffi/src/mock_server/handles.rs#L1707-L1729

I would need to enable trace-level logging and run the Pact PHP tests to properly understand that is going on.

PR #459, opened 2024-08-05
feat: allow non-object matchers for array-contains

MaxCampman:
I have recently been writing a lot of pact tests and I noticed that the array-contains matcher requires each variant to be an object.

Reading the spec, it says "Checks if all the variants are present in an array".
* https://github.com/pact-foundation/pact-specification/tree/version-4

Based on this, and the below implementation guides, it seems that non-object types should be supported too.
* https://docs.pact.io/implementation_guides/javascript/docs/matching#array-contains-matcher
* https://docs.pact.io/implementation_guides/jvm/consumer#array-contains-matcher-v4-specification

Having given the code base a cursory review, it seems that supporting this should be quite trivial.

Note: this is my first PR here so please let me know if there are steps I have missed

### DIFF
--- a/rust/pact_ffi/src/mock_server/bodies.rs
+++ b/rust/pact_ffi/src/mock_server/bodies.rs
@@ -8,7 +8,7 @@ use either::Either;
 use lazy_static::lazy_static;
 use regex::Regex;
 use serde_json::{Map, Value};
-use tracing::{debug, error, trace, warn};
+use tracing::{debug, error, trace};
 
 use pact_models::bodies::OptionalBody;
 use pact_models::content_types::ContentTypeHint;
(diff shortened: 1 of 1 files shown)


rholshausen:
While this is a trivial change, I'm unsure that it would work correctly. If the variant is an array (like `[A, B]`), I think that means that the source array must include a sub-array with the values `A, B` in that order.

Is this what you require?

MaxCampman:
> While this is a trivial change, I'm unsure that it would work correctly. If the variant is an array (like `[A, B]`), I think that means that the source array must include a sub-array with the values `A, B` in that order.
> 
> Is this what you require?

This seems to be consistent with the rest of pact. If I wanted to check that a sub-array contained certain items in no particular order, I would be able to specify further "array-contains" matchers in a nested fashion, something like:

```csharp
Match.ArrayContains([
    Match.ArrayContains(["A", "B"]),
    Match.ArrayContains(["C", "D"]),
])
```

As you say, if I specified just the arrays, they would need to match in the specified order:

```csharp
Match.ArrayContains([
    new[] { "A", "B" },
    new[] { "C", "D" },
])
```

Not sure what kind of use case would require nested array-includes, but nonetheless I think this change makes it possible to specify both restrictions

Can see [this as a test case](https://github.com/MaxCampman/pact-net/blob/aa5bd694e8be0a563bd735254c8ffc0cc7e363b2/tests/PactNet.Tests/Matchers/ArrayContainsMatcherTests.cs#L170) I added to a fork of PactNet running on top of the changes in this PR

rholshausen:
Awesome, I'll merge it.

PR #461, opened 2024-08-30
fix(ffi): annotate body in raw body branch

JP-Ellis:
If the content type was not otherwise handled,
`OptionalBody::from(&str)` was used. This sets the content type associated with the body to `None` irrespective of the content type passed through to `process_body`.

This fixes the issue by ensuring the `OptionalBody::Present` is created with the appropriate content type.

### DIFF
--- a/rust/pact_ffi/src/mock_server/handles.rs
+++ b/rust/pact_ffi/src/mock_server/handles.rs
@@ -1730,7 +1730,11 @@ fn process_body(
         _ => {
             // We either have no content type, or an unsupported content type.
             trace!("Raw body");
-            OptionalBody::from(body)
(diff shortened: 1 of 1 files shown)


tienvx:
Should we add a simple test to verify if it work and make sure this issue will not happen in the future?

Something like this:

```
pactffi_with_body('a=1&b=2&c=3&d=4', 'x-www-form-urlencoded')
assert_body('a=1&b=2&c=3&d=4',)
assert_header_is_set('Content-Type', 'x-www-form-urlencoded')
```

rholshausen (approved):

tienvx (approved):

PR #474, opened 2024-11-26
fix(verifier): add Accept-Charset=utf8 header when post/put

YOU54F:
Fixes 406 Not Acceptable on publishing verification results, when using a pact broker started with thin. (puma / webrick do not exhibit the issue


YOU54F:
there is an issue with the compat suite deps not introduced by this change

Issue #471, opened 2024-10-30
Examples - Remove language implementations which show usage of the FFI 


YOU54F:
Remove

https://github.com/pact-foundation/pact-reference?tab=readme-ov-file#other-languages

as these have been implemented natively in respective languages, provide links to

PR #475, opened 2024-11-26
Chore/remove stale examples

YOU54F:
fixes https://github.com/pact-foundation/pact-reference/issues/471

Cleans up codebase a bit which uses outdated examples, which are better served by pointing to the latest implemented client dsl's


JP-Ellis:
Yay! Looks great!

And with keeping just C, it makes it a lot clearer what the intended call order ought to be without the overhead of each languages' specific FFI quirks.

YOU54F:
Agreed, less is more when trying to diagnose things over system boundaries, less to go stale, less to maintain and - single place to centralise best practise usage and arrangement of the pact ffi calls.

for end users it just served as a source of confusion and aren’t particularly helpful anymore especially when we have implementations much further along than the demos.

for anyone else as an intro to the pact ffi, the test suite there is probably the best place to start as there are existing tests that can be modified to suit :)

possibly good for a repro template if users having ffi issues

JP-Ellis (approved):

PR #480, opened 2024-12-11
chore: reduce ci usage

JP-Ellis:
This should help reduce the CI usage by:

1. Only running  CI on either:

   - PRs targetting `master`
   - pushes to `master`

   Most importantly, CI is no longer doubled in PRs (as previously, `push` and `pull_request` would _both_ be triggered, consuming double the number of runners).

2. Cancelling existing runs if a new commit is added

3. Speeding up some existing builds by not unnecessarily rebuilding everything (especially with the compatibility suite)

4. Using nextest in place of `cargo test` as it is typically faster

5. Using caching of build dependencies where possible

### DIFF
--- a/.github/workflows/build-ffi.yml
+++ b/.github/workflows/build-ffi.yml
@@ -33,13 +33,10 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
         with:
           toolchain: stable
+          components: rustfmt
       - name: Install shared mime info DB
         if: runner.os == 'macOS'
         run: brew install shared-mime-info
-      - uses: dtolnay/rust-toolchain@nightly
-        with:
-          toolchain: nightly
-          components: rustfmt
       - name: Install doxygen
         if: runner.os == 'Linux'
         run: sudo apt-get install -y doxygen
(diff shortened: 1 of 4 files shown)


.github/workflows/build-ffi.yml
```diff
@@ -1,6 +1,22 @@
...
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref || github.run_id }}
+  cancel-in-progress: true
```
rholshausen, line 13: I'll remove that, that would be annoying. Imagine waiting for a build, then someone else pushes a commit.

JP-Ellis: The concurrency group is per-PR

JP-Ellis: So there should not be a case of waiting for a build, and it being canceled by someone else's push _except_ if you are both working on the same PR.

If you want this to run specifically when creating tagged released, I recommend creating a separate trigger for tags.

PR #486, opened 2025-02-10
feat: Implement generators for XML

tienvx:


tienvx:
This step is failing on multiple checks:

> Upload Release Assets
> Run svenstaro/upload-release-action@v2
> Error: Resource not accessible by integration - https://docs.github.com/rest/releases/releases#create-a-release

I think we should disable this step on pull requests. @uglyog

rholshausen:
I don't know why it is running the release build for your PR, I did not so that.

rholshausen:
The only issue I can see with this, is it walks the entire XML tree for every generator. For large XML documents, this would be inefficient. I think that the `generate_values_for_xml_element` could return if the current node does not match part of the generator path, or the current recursion depth is greater than the length of the generator path.

I.e. if the generator path is `$.parent.child`, then when checking the `child` nodes, no point is checking their children.

Can also optimise for attributes and text nodes. For those, the generator path has to end with either `@name` or `#text`. If it doesn't, no need to check those.

YOU54F:
> This step is failing on multiple checks:
> 
> > Upload Release Assets
> > Run svenstaro/upload-release-action@v2
> > Error: Resource not accessible by integration - https://docs.github.com/rest/releases/releases#create-a-release
> 
> I think we should disable this step on pull requests. @rholshausen

agreed, previously the upload step was only setup to run for release tags

https://github.com/pact-foundation/pact-reference/blob/e2273a9a83c8d6b175e1295167e27172657040c9/.github/workflows/release.yml#L119

although we had some CI optimisations that broke the release workflow, so we did a revert, and I think some things got missed.

It would still be nice to get the ci optimisations in, however scoping the publish assets step to release tags only, would probably be better?

YOU54F:
> I don't know why it is running the release build for your PR, I did not do that.

it was setup to smoke test the release build, prior to merging to master, doing the release and finding out it doesn't build. (it the only place we build all the cross platform binaries). it should be scoped to only upload on release though ideally.

YOU54F:
> I.e. if the generator path is `$.parent.child`, then when checking the `child` nodes, no point is checking their children.
> 
> Can also optimise for attributes and text nodes. For those, the generator path has to end with either `@name` or `#text`. If it doesn't, no need to check those.

Excellent suggestions!

tienvx:
> The release pipeline now also runs on PRs. If being run on PRs, a
  non-release build done as it is faster.

I think the `Release workflow` is run on PRs on [purpose](https://github.com/pact-foundation/pact-reference/commit/7732a637b0e890b9ecb29470a03908d8f30a523a). Can you help me fix these failing checks @JP-Ellis ? I think it's related to permission issue.

tienvx:
I refactored the code to implement the suggested performance improvements.

rholshausen:
This PR is still in draft mode?

tienvx:
It is ready for review, there are some notes:
* I need help fixing failing checks
* I also need reviewing RFC PR too https://github.com/pact-foundation/roadmap/pull/113

tienvx:
I fixed the failed checks. Please review again @rholshausen . Be aware of #492

rholshausen:
I know of #492, I just don't know why it happens, but it feels like a typical race condition.

PR #487, opened 2025-02-11
feat: Use new XML generator

tienvx:
Depend on #486


tienvx:
This MR is ready for review @rholshausen

PR #489, opened 2025-03-13
feat(test):add tests for different date matcher scenarios in pact_ffi

pulak777:
Tests for different `date` matcher scenarios  (applicable for datetime/time as well) - 
1. when `pact:generator:type` is provided in the body matcher, mock server returns current time (Local::now()) in the provided format
2. when not provided, mock server returns the specified value

Context - Currently when using pact-js v3 matcher type `date` and `time`, mock-server returns current date/time while it's expected to return the example date/time. The issue is not originating from here (will create a seperate pr for that in pact-js), but these additional tests might worth having.

### DIFF
--- a/rust/pact_ffi/tests/tests.rs
+++ b/rust/pact_ffi/tests/tests.rs
@@ -9,6 +9,7 @@ use std::str::from_utf8;
 use std::time::Duration;
 
 use bytes::Bytes;
+use chrono::Local;
 use expectest::prelude::*;
 use itertools::Itertools;
 use libc::c_char;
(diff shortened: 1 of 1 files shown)


rholshausen:
There is a compile error with the tests:
```
error[E0433]: failed to resolve: use of undeclared type `Local`
    --> pact_ffi/tests/tests.rs:2216:24
     |
2216 |             let date = Local::now().format("%Y-%m-%d").to_string();
     |                        ^^^^^ use of undeclared type `Local`
     |
help: consider importing this struct
     |
1    + use chrono::Local;
     |
```

PR #504, opened 2025-09-05
fix: query param ordering respected when creating v2 pacts

YOU54F:
In creating test for the pact_broker-client rewrite in rust, I need to coerce the Pact files to V2.

This gives me issues with the ordering of certain query string parameters which are expected after the specified participant.

This PR looks to introduce a failing test, and a fix for the issue

### DIFF
--- a/rust/pact_models/src/query_strings.rs
+++ b/rust/pact_models/src/query_strings.rs
@@ -311,4 +311,67 @@ use crate::query_strings::build_query_string;
     assert_eq!(result, "pacticipant=foo&version=1%2e2%2e3&pacticipant=bar&version=4%2e5%2e6");
   }
 
+
+    // sends URL Querying broker at: http://127.0.0.1:61567/matrix?q[][pacticipant]=Foo&q[][version]=1.2.3&q[][pacticipant]=Bar&q[][latest]=true&q[][tag]=prod&latestby=cvpv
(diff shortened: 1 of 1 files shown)


YOU54F:
Hmm still not quite quite

```
                        .path("/matrix")
                        .query_param("q[][pacticipant]", "Foo")
                        .query_param("q[][version]", "1.2.3")
                        .query_param("q[][pacticipant]", "Bar")
                        .query_param("q[][latest]", "true")
                        .query_param("q[][tag]", "prod")
                        .query_param("latestby", "cvpv");
```

should return ordering like

```
query: "q%5B%5D%5Bpacticipant%5D=Foo&q%5B%5D%5Bversion%5D=1.2.3&q%5B%5D%5Bpacticipant%5D=Bar&q%5B%5D%5Blatest%5D=true&q%5B%5D%5Btag%5D=prod&latestby=cvpv"
```

but returns

```
"query": "latestby=cvpv&q[][latest]=true&q[][pacticipant]=Foo&q[][tag]=prod&q[][version]=1%2e2%2e3&q[][pacticipant]=Bar"
```

I'll just overwrite the headers in the pact broker verification task instead

rholshausen:
See https://stackoverflow.com/questions/30243100/how-do-i-sort-a-map-by-order-of-insertion

BTW, changing the map type would be a breaking change as those structs are used everywhere.

PR #509, opened 2025-12-15
feat(ffi): upgrade mock server to v2

JP-Ellis:
Upgrade the mock server dependency within the FFI crate to version 2.

In general, this should be backwards compatible, and should bring through to the FFI a few fixes.

### DIFF
--- a/rust/pact_ffi/Cargo.toml
+++ b/rust/pact_ffi/Cargo.toml
@@ -33,7 +33,7 @@ onig = { version = "6.5.1", default-features = false }
 # FFI Inmemory sink uses the global static `LOG_BUFFER` in the pact_matching crate. This is required to be the same
 # version as the mock server crate
 pact_matching = { version = "1.2.12" }
-pact_mock_server = { version = "~1.2.16" }
+pact_mock_server = "~2.1.0"
 pact_models = { version = "~1.3.7" }
 pact-plugin-driver = { version = "~0.7.4" }
 pact_verifier = { version = "~1.3.2", path = "../pact_verifier" }
--- a/rust/pact_ffi/src/mock_server/mod.rs
+++ b/rust/pact_ffi/src/mock_server/mod.rs
@@ -43,31 +43,36 @@
 
 #![warn(missing_docs)]
 
-use std::{ptr, str};
 use std::any::Any;
-use std::ffi::CStr;
-use std::ffi::CString;
+use std::ffi::{CStr, CString};
 use std::net::ToSocketAddrs;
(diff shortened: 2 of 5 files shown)


rust/pact_ffi/Cargo.toml
```diff
@@ -33,7 +33,7 @@ onig = { version = "6.5.1", default-features = false }
 # FFI Inmemory sink uses the global static `LOG_BUFFER` in the pact_matching crate. This is required to be the same
 # version as the mock server crate
 pact_matching = { version = "1.2.12" }
-pact_mock_server = { version = "~1.2.16" }
+pact_mock_server = "2.1.3"
```
rholshausen: Set the version to `~2.1.3` please, so it will only automatically update patch versions

rust/pact_ffi/src/mock_server/mod.rs
```diff
@@ -183,192 +212,267 @@ ffi_fn! {
...
-    Err(cause) => {
-      error!("Caught a general panic: {:?}", cause);
-      false
+ffi_fn! {
+    /// External interface to check if a mock server has matched all its requests. The port number is
+    /// passed in, and if all requests have been matched, true is returned. False is returned if there
+    /// is no mock server on the given port, or if any request has not been successfully matched, or
+    /// the method panics.
+    #[tracing::instrument(level = "trace")]
+    fn pactffi_mock_server_matched(mock_server_port: i32) -> bool {
```
rholshausen: I think we still need the catch_unwinds here, because, for instance, this line
```
let mut guard = MANAGER.lock().unwrap();
```
will panic if the mutex has been poisoned, and panics across FFI boundaries result in undefined behaviour.

rholshausen:
No idea why the builds are failing. I tried rerunning them, but it might be they are taking too long on the MacOS agents that the agents are being killed. The v2-engine might be some regression, but your changes should not be affecting that.

rust/pact_ffi/Cargo.toml
JP-Ellis: Cargo will automatically upgrade patch versions without the `~`, under the [default requirements syntax](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#default-requirements).

rholshausen: > Tilde requirements specify a minimal version with some ability to update. If you specify a major, minor, and patch version or only a major and minor version, only patch-level changes are allowed.

From that link

JP-Ellis: All good, I missed the 'only'

PR #524, opened 2026-04-28
fix: remove panic on invalid content type

JP-Ellis:
If the content type is invalid, an `unwrap` caused the `pactffi_with_body` function to panic. The panic has been replaced by a fallback to log an error, and fallback to auto-detection of the payload based on the body itself.

Ref: pact-foundation/pact-js#1754
Ref: PACT-6546


rholshausen (approved):

JP-Ellis:
Both Copilot comments were valid and have been addressed.

PR #529, opened 2026-04-30
Feat: HTML verification report

rholshausen:
Adds an `--html` option to the verifier CLI to render an HTML report of the verification.

Sample report:
image


mefellows:
This is great! You know what would be _really_ helpful to add - output logs in collapsable or perhaps in a separate page in a directory. This way you can isolate the logs for the scenario.

But that shouldn't stop this going out. If we wanted to future proof it, we could specify a directory for the `--html` option, so over time we could add multiple supporting artifacts.

PR #534, opened 2026-05-26
feat: add --retries / PACT_BROKER_HTTP_RETRIES to pact_verifier_cli

JP-Ellis:
Exposes broker HTTP retry count as --retries (env: PACT_BROKER_HTTP_RETRIES,
default: 8) in pact_verifier_cli, matching the flag in pact-broker-cli.

The retry count is threaded through VerificationOptions.broker_request_retries into the HALClient used for all broker calls: pact fetching (both legacy and dynamic) and publishing verification results.

pact_verifier and cli bumped to 1.4.0

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2976,7 +2976,7 @@ dependencies = [
 
 [[package]]
 name = "pact_verifier"
-version = "1.3.6"
+version = "1.4.0"
 dependencies = [
  "ansi_term",
  "anyhow",
@@ -3038,7 +3038,6 @@ dependencies = [
  "pact_verifier",
  "quick-xml 0.39.2",
  "regex",
- "reqwest 0.12.28",
  "rstest 0.24.0",
  "serde",
  "serde_json",
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -19,13 +19,13 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "aes"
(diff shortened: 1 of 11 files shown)


rholshausen:
> pact_verifier and cli bumped to 1.4.0

Such a tiny change for a version bump?

JP-Ellis:
The main thing here is a new flag, hence the new minor version.

Technically, there's a breaking change (the typo fix is part of the public interface...).

Up to you what version bump you prefer. I can fix up this PR easily.

rholshausen:
Ok, I've reset the version on the CLI but kept the version bump for the verifier crate.

rholshausen:
There is a failing test, as the error text has now changed:
```
'pact_broker::tests::fetch_returns_an_error_if_it_does_not_get_a_valid_hal_response' (10134) panicked at pact_verifier/src/pact_broker.rs:1365:9:
assertion failed: `(left == right)`

Diff < left / right > :
<Error with the content of a HAL resource - Did not get a valid HAL response body from pact broker path '/nonhal' - error decoding response body. URL: 'http://127.0.0.1:43399/'
>Error with the content of a HAL resource - Did not get a valid HAL response body from pact broker path '/nonhal' - error decoding response body for url (
http://127.0.0.1:43399/nonhal
)
. URL: 'http://127.0.0.1:43399/'
```

JP-Ellis:
Couldn't be bothered dealing with the merge commit when I rebased... so it clobbered the git history. Anyway, error is fixed, kept the minor version bump for the `pact_verifier` breaking change, and reverted the version bump for the `pact_verifier_cli`.

PR #549, opened 2026-09-07
feat(pact_verifier): include pact and mismatch details in interaction results

frudisch:
## Summary

The JSON verification result (`pactffi_verifier_json`, `pact-verifier --json`) lists one entry per verified interaction in `interactionResults`, but each entry only carries `description`, `result` and `duration`. Consumer and provider are only available as free text in `errors[].interaction`, provider states not at all, and mismatches have to be joined to the interaction via that text.

When several pacts of different consumers are verified in one run, the entries cannot be assigned to their pact. This prevents consumers of the JSON (language bindings such as pact-js, or test framework reporters) from reporting the outcome per pact and per interaction, for example as one test suite per pact with one test case per interaction.

## Changes

- `VerificationInteractionResult` gains `consumer`, `provider` and `provider_states` (the state names in pact order). They are populated in `verify_pact_internal`, where the pact and interaction are available.
- Every `interactionResults` entry in the JSON additionally has:
  - `consumer`, `provider`
  - `providerStates` (array of names, empty if none)
  - `pending`
  - `mismatch` for failed interactions, with the same structure as `errors[].mismatch`
- Existing attributes (`description`, `result`, `duration`, `interactionId`, `interactionKey`) are unchanged, so the change is additive. `errors`, `pendingErrors`, `output` and the JUnit XML report are unchanged.
- Documents the structure of the JSON report in the `pact_verifier_cli` README / crate docs and in the doc comment of `pactffi_verifier_json`.

Example entry for a failed interaction:

```json
{
  "consumer": "DemoConsumer",
  "provider": "DemoProvider",
  "description": "interaction 3 - GET /three",
  "providerStates": ["resource three is available"],
  "pending": false,
  "result": "Error",
  "mismatch": {
    "type": "mismatches",
    "interactionId": "",
    "mismatches": [
      { "type": "BodyMismatch", "path": "$.ok", "expected": "true", "actual": "\"true\"", "mismatch": "Expected 'true' (String) to be the same type as true (Boolean)" }
    ]
  },
  "duration": "2ms"
}
```

## Test plan

- [x] `cargo test -p pact_verifier` (lib and integration tests), including a new integration test `interaction_results_include_the_pact_and_provider_state_details` with a pact fixture that has interactions with and without provider states
- [x] Extended `verification_execution_result_to_json_includes_interaction_details` to cover the new attributes and the `mismatch` attribute for both the error and the mismatches variant
- [x] `cargo test -p pact_verifier_cli --lib reports` (JUnit XML snapshots unchanged)
- [x] `cargo check -p pact_ffi`, `cargo doc -p pact_verifier_cli --no-deps`
- [x] End-to-end: built `pact-verifier` and verified two pact files of different consumers against a passing and a deliberately failing provider with `--json`; every entry carries its consumer, provider, provider states and, when failed, its mismatches

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/rust/pact_ffi/src/verifier/mod.rs
+++ b/rust/pact_ffi/src/verifier/mod.rs
@@ -702,6 +702,14 @@ ffi_fn! {
     /// Extracts the verification result as a JSON document. The returned string will need to be
     /// freed with the `free_string` function call to avoid leaking memory.
     ///
+    /// The document has the attributes `result` (overall pass/fail), `notices`, `output` (the
+    /// verification output lines), `errors`, `pendingErrors` and `interactionResults`. Each entry
+    /// in `interactionResults` has the attributes `consumer`, `provider`, `description`,
+    /// `providerStates`, `result` (`OK` or `Error`), `pending`, `duration`, a `mismatch` attribute
+    /// with the details when the result is `Error`, and `interactionId`/`interactionKey` when
+    /// they are known. See the JSON report section in the pact_verifier_cli documentation for
+    /// more details.
+    ///
     /// Will return a NULL pointer if the handle is invalid.
     fn pactffi_verifier_json(handle: *const handle::VerifierHandle) -> *const c_char {
       let handle = as_ref!(handle);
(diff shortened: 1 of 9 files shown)


mefellows:
Nice idea, thank you!

PR #553, opened 2026-09-21
feat(ffi): expose a log callback for host language logging

JP-Ellis:
## Goal

Give host languages a way to receive every log event from the Pact libraries as it happens, through a C callback they can route into their own logging framework, and make plugin log delivery work regardless of how logging was initialised.

Today the FFI logger can only write to stdout, stderr, a file or an in-memory buffer. The buffer loses each event's level and target, and none of the sinks tell the host which test or mock server an event belongs to. Separately, the plugin log sink added in 0.5.5 was only installed by `pactffi_init` / `pactffi_init_with_log_level`, so a host configuring logging through `pactffi_logger_apply` or `pactffi_log_to_*` never received a plugin log entry, and `pactffi_register_plugin_log_callback(NULL)` did not deregister despite its documentation saying it would.

## Sequence

1. **Install the plugin log sink on callback registration.** `pactffi_register_plugin_log_callback` and `pactffi_logger_apply` now install the sink (a `OnceLock`, so repeated calls are harmless), and passing `NULL` deregisters the callback.
2. **Add the callback layer.** A `tracing_subscriber::Layer` forwards each event at or below a configurable level to a registered C callback, with the ID of the emitting operation (mock server ID, or `verify:<provider>` during verification), the test run ID for the emitting thread, the level, the target and the message. The argument order matches `PluginLogCallback`.
3. **Expose it.** `pactffi_register_log_callback` and `pactffi_set_log_callback_level` are added, and the layer is attached to every subscriber the FFI installs. The subscriber admits the more verbose of the sink levels and the callback level; each sink filters at its own level, so raising the callback level does not change what reaches stderr or a file.

## Alternatives weighed

- Forwarding through the existing buffer sink would have kept the API surface smaller, but the buffer discards level and target and has no per-thread attribution, which is the information a host logging framework needs.
- Installing the plugin sink from `pactffi_register_plugin_log_callback` alone would have fixed the documented behaviour but still left `pactffi_logger_apply` users without plugin logs, so both entry points install it.
- With `pactffi_init`, the environment filter bounds what the callback receives, since the `EnvFilter` is the subscriber's global filter. This is documented on `pactffi_set_log_callback_level` instead of being worked around.

## Verification

- `cargo test --lib` in `rust/pact_ffi`: 209 passed. New tests cover level filtering and field delivery through the layer, the level round trip through the FFI, `NULL` deregistration through the public `pactffi_register_plugin_log_callback`, and the sink/callback level combination.
- End to end from pact-python against a locally built library: events arrive through the callback with the test run ID attached, and plugin log entries are delivered without calling `pactffi_init`.

## Downstream

pact-python's bindings are being updated alongside this (pact-foundation/pact-python#1694) and depend on the 0.5.9 release.


rholshausen:
LGTM

I have been avoiding adding callbacks because an invalid callback pointer will cause a segment violation or other undefined behavior. I think this is special case where the benefit out-ways the risk.

JP-Ellis:
Agreed, callbacks can be a pain; but in this case, a one-way 'fire and forget' should hopefully not cause too many opportunities for violating the various guarantees Rust expects.

Thanks for the review!

